PR #19, opened 2023-03-22
Upload Button/Progress Bar/Init bootloader process

josemapac2:
+ Adding a progress bar close to the button "Upload" in the "Device Settings/Update Firmware" menu.
+ Making the "Upload" button work to Update the Firmware, by forcing the bootloader process sending a command through the USB serial port. 
+ Fix the bug in the view CONNECT DEVICE to automatically detect all Daqifi USB ports: Check there is no 'null'


Daqifi.Desktop.Setup/UpgradeLog2.htm
```diff
@@ -0,0 +1,268 @@
+﻿<!DOCTYPE html>
```
tylerkron, line 1: We don't need to check in this file

Daqifi.Desktop.Setup/UpgradeLog.htm
```diff
@@ -0,0 +1,278 @@
+﻿<!DOCTYPE html>
```
tylerkron, line 1: Don't need to check in this file

Daqifi.Desktop/ViewModels/DaqifiViewModel.cs
```diff
@@ -48,6 +56,227 @@ public class DaqifiViewModel : ViewModelBase
         private bool _canToggleLogging;
         private string _loggedDataBusyReason;
         private string _firmwareFilePath;
+        //private readonly Pic32Bootloader _bootloader;
```
tylerkron, line 59: Delete this commented line?

Daqifi.Desktop/ViewModels/DaqifiViewModel.cs
```diff
@@ -16,9 +18,15 @@
 using System.Collections.ObjectModel;
 using System.ComponentModel;
 using System.Linq;
+using System.IO;
 using System.Threading.Tasks;
 using System.Windows;
 using System.Windows.Input;
+using System.Collections;
+using System.Diagnostics;
```
tylerkron, line 26: This is unused

Daqifi.Desktop/ViewModels/DaqifiViewModel.cs
```diff
@@ -16,9 +18,15 @@
...
+using Daqifi.Desktop.Device.SerialDevice;
+using System.Threading;
+using System.Windows.Interop;
```
tylerkron, line 29: This is unused

tylerkron (changes requested):
Thanks for the PR! Just a couple of cleanup items I saw!

cptkoolbeenz (approved):
Using firmware update feature with stock bootloader causes Windows confusion when switching from COM to HID device rapidly.  Adding a 2s delay during bootloader startup solves this issue.

Issue #38, opened 2024-03-16
Check for new versions of the application


tylerkron:
As a user, I want to know if there are new versions of this app. I should see a notification on startup if my version is out of data and a link to download the new version (hosted on GitHub).

PR #59, opened 2024-10-07
feat: #38 check for new versions of the application

codedecoded01723:
1. Issue resolved for new save profile after profile file delete or profiles delete  .
2. Multiple profile selection removed .
3. slider frequency issue in add profile resolved .
4. Added new flyouts for notification .
5. implemented logics for the version update notifications .
6. codacy issues resolved


tylerkron (approved):
There might be issues with adding multiple channels, but we'll merge this and make tickets for any issues that we find.

Issue #51, opened 2024-10-01
Add channels from 2+ devices


tylerkron:
Right now there is a limitation on the channel name. For example if I add `DIO1` from one device, I can't add `DIO1` from a second device.

A channel doesn't currently know what device it belongs to. Add a device name to the `IChannel` interface so that we can add, log, and export channels and know what device that channel came from.

## Success Criteria

- As a user I can add the same channel name from two devices
- As a user I can stream the same channel name from two devices
- As a user I can export the same channel name from two devices

PR #72, opened 2024-12-10
feat: Add channels from two device(#51)

codedecoded01723:
feat:
1.Add two device 
2.Add channel from two device 
3.Save settings of two device and ,
4.Profile activation supports two device


tylerkron (changes requested):
Hey @codedecoded01723

Some of this works and some doesn't. I can add multiple devices, configure device settings, add channels and save that as a profile. All of that works great. However, I can't actually stream with two devices.

For example, I added a real DAQiFi device and added a simulator device. Only one device's data gets shown / logged. Is this expected and you'd handle streaming in another ticket. or should this work as part of this ticket?

Picture as an example:
image

tylerkron:
Only 1 device is still getting captured on the live graph and logged data.

To reproduce:

1. Connect a physical device via WiFi
2. Connect a simulated device via WiFi
3. Set both devices to 10Hz
4. Add A0 from both devices.
5. Turn on the logging session.
6. Notice that only device's data is showing up on the live graph
7. Stop the logging session.
8. Notice that the logged session only has one device's worth of data

tylerkron (approved):

PR #73, opened 2024-12-11
chore: testing out basic ci

tylerkron:


tylerkron:
This might require a windows runner still

PR #87, opened 2025-01-06
feat: Added option to export relative time vs absolute time

codedecoded01723:


tylerkron (changes requested):
This isn't working as expected. It's truncating (or rounding down) to the whole second. For example if I am streaming at 100Hz I'd expect the time increments for the relative export to look like:

```
0.000
0.010
0.020
0.030
....
```

However if you try this, you get `0.000` until it's been a whole second and then you get `1.000` until it's been two seconds.

Attached is a comparison of a logging session exported with absolute and relative formats.

[absolute.csv](https://github.com/user-attachments/files/18370456/absolute.csv)
[relative.csv](https://github.com/user-attachments/files/18370457/relative.csv)

PR #92, opened 2025-01-17
fix: Relsolved the renaming a logging session

codedecoded01723:


tylerkron (changes requested):
This only resolves the first acceptance criteria:

```
Renaming a logging session persists across application restarts.
Exported logging session files use the updated session name as the default file name.
```

The exported log session does not have the updated names. See attached screenshots of when I did an export all

image

image

PR #173, opened 2025-07-21
fix: resolve UDP discovery and TCP connection issues in multi-NIC environments

cptkoolbeenz:
## Summary
- Fixes UDP device discovery not working with WiFi adapters
- Resolves TCP connection timeouts when multiple NICs have routes to same IP
- Adds backward compatibility for firmware that responds to port 30303

## Problem
1. UDP discovery broadcasts were only sent from Ethernet interfaces, not WiFi
2. When multiple network interfaces had routes to 192.168.1.1 (common for routers and DAQiFi AP mode), Windows would route TCP connections to the wrong interface
3. Device discovery failed in AP mode

## Solution
1. Created `DaqifiDeviceFinderMultiInterface` to broadcast from ALL network interfaces
2. Created `DaqifiDeviceFinderBackwardCompatible` that:
   - Listens on port 30303 for existing firmware responses
   - Tracks which local interface discovered each device
3. Modified `DaqifiStreamingDevice` to bind TCP connections to the correct local interface

## Test plan
- [x] Test UDP discovery with device in AP mode
- [x] Test UDP discovery with device in STA mode  
- [x] Verify device discovery works from WiFi adapters
- [x] Confirm TCP connections succeed when router and device share same IP
- [ ] Test with multiple DAQiFi devices on different interfaces
- [ ] Verify backward compatibility with older firmware versions

## Breaking changes
None - maintains full backward compatibility

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/Daqifi.Desktop/Device/AbstractStreamingDevice.cs
+++ b/Daqifi.Desktop/Device/AbstractStreamingDevice.cs
@@ -13,6 +13,7 @@ using Daqifi.Desktop.IO.Messages.Producers;
 using ScpiMessageProducer = Daqifi.Core.Communication.Producers.ScpiMessageProducer;
 using System.Runtime.InteropServices; // Added for P/Invoke
 using CommunityToolkit.Mvvm.ComponentModel; // Added using
+using System.Threading;
 
 namespace Daqifi.Desktop.Device;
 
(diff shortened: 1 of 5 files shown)


cptkoolbeenz:
Fixed feature regression from merge(s).  Tested on multiple NICs.

PR #457, opened 2026-04-04
fix: improve chart performance with large datasets using LTTB downsampling

### LINKED ISSUE #36: Slow Display With Large Dataset
When collecting a “large” set of samples (for example 100,000), the interactive display of logged data gets extremely slow, to the point of not being useable.

tylerkron:
## Summary

Closes #36

- Implements the **Largest-Triangle-Three-Buckets (LTTB)** downsampling algorithm to reduce rendered data points to ~5,000 per channel while preserving visual shape (peaks, troughs, spikes)
- **Live chart (PlotLogger)**: Increases raw data buffer from 5k to 50k points per channel (10x more history), decimates to 5k for rendering every 1s cycle. Uses gap-aware decimation to preserve line breaks at data stream gaps
- **Stored session playback (DatabaseLogger)**: Replaces the hard 1M point cap (which silently dropped data) with per-channel LTTB decimation that displays all data. Adds `OrderBy(TimestampTicks)` to ensure time-sorted data for correct decimation
- Adds `DataPointDecimator` with both standard and gap-aware decimation methods
- Adds 15 unit tests covering null/edge cases, output size, first/last point preservation, visual fidelity (sine wave peaks, spike preservation), ordering, gap marker preservation, and performance (1M points in <1s)

## Test plan

- [ ] Verify live chart performance with high-frequency streaming (multiple channels)
- [ ] Verify stored session playback with 100k+ sample sessions renders quickly
- [ ] Verify gap markers (line breaks) are preserved in live chart when device disconnects/reconnects
- [ ] Verify chart subtitle shows "Showing X downsampled points from Y total samples" for large sessions
- [ ] Verify visual fidelity — peaks and spikes should still be visible after downsampling
- [ ] Run unit tests on Windows CI

### DIFF
--- a/Daqifi.Desktop.Test/Loggers/CircularBufferTests.cs
+++ b/Daqifi.Desktop.Test/Loggers/CircularBufferTests.cs
@@ -1,0 +1,141 @@
+using Daqifi.Desktop.Logger;
+
+namespace Daqifi.Desktop.Test.Loggers;
+
+[TestClass]
+public class CircularBufferTests
+{
+    #region Constructor Tests
+    [TestMethod]
+    public void Constructor_ZeroCapacity_Throws()
+    {
+        Assert.ThrowsExactly<ArgumentOutOfRangeException>(() => new CircularBuffer<int>(0));
+    }
+
+    [TestMethod]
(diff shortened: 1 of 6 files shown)


tylerkron:
will come up with a better solution

PR #491, opened 2026-04-18
fix: Connect Device dialog close race + empty-selection close

tylerkron:
## Summary
- Fixes two pre-existing bugs in the Connect Device dialog flagged by Qodo on [#488](https://github.com/daqifi/daqifi-desktop/pull/488) and scoped out of that visual redesign. Both date back to the SVN migration (`ec6602d`).
- **Close race:** Connect buttons wired both `Command="{Binding …Command}"` (async) and `Click="btnConnect_Click"` (sync `Close()`). The window closed while the connect task was still running; the `Closing` handler then ran `vm.Close()` which disposed discovery concurrently with the in-flight connect.
- **Connect closes with no selection:** Connect buttons were always enabled — clicking with an empty `SelectedItems` list no-op'd on the VM side but still closed the dialog, so the user thought they'd connected when they hadn't.

## Approach
- `ConnectionDialogViewModel` raises a new `CloseRequested` event **only after** a connect command's awaited work completes successfully. Empty input or failed connect → no event → dialog stays open.
- `ConnectionDialog.xaml.cs` subscribes to `CloseRequested` and marshals to the UI thread before closing the window.
- Removed `Click="btnConnect_Click"` from the 4 action buttons (WiFi, Manual WiFi, USB, Manual USB). The Firmware button was already a no-close case.
- `ConnectionDialogViewModel.Close` is now idempotent — double-invocation (once from the connect path, once from the Closing handler) can't double-dispose discovery finders.
- Added a `ToStreamingDevices` helper that safely casts/lists `SelectedItems` and makes the empty-check explicit.

## Test plan
- [ ] Happy path: pick a WiFi/USB device, click Connect — dialog closes after the VM's connect completes, device shows up connected.
- [ ] Empty selection: click Connect on WiFi tab with no tiles selected — dialog stays open.
- [ ] Manual WiFi / Manual USB: blank field + Connect → dialog stays open; populated field → connects and closes.
- [ ] Firmware tab: select HID device → Firmware dialog opens without closing the Connect dialog (unchanged behavior).
- [ ] Close via X during a slow connect — no double-dispose exceptions in the NLog output at `%CommonApplicationData%\DAQifi\Logs`.
- [ ] New unit tests under `Daqifi.Desktop.Test/ViewModels/ConnectionDialogViewModelCloseTests.cs` pass on Windows.

### DIFF
--- a/Daqifi.Desktop.Test/ViewModels/ConnectionDialogViewModelCloseTests.cs
+++ b/Daqifi.Desktop.Test/ViewModels/ConnectionDialogViewModelCloseTests.cs
@@ -1,0 +1,119 @@
+using Daqifi.Desktop;
+using Daqifi.Desktop.Device.SerialDevice;
+using Daqifi.Desktop.DialogService;
+using Daqifi.Desktop.ViewModels;
+using Moq;
+
+namespace Daqifi.Desktop.Test.ViewModels;
+
+[TestClass]
+public class ConnectionDialogViewModelCloseTests
+{
+    private Func<DuplicateDeviceCheckResult, DuplicateDeviceAction>? _originalDuplicateDeviceHandler;
+
+    [TestInitialize]
+    public void TestInitialize()
+    {
+        _originalDuplicateDeviceHandler = ConnectionManager.Instance.DuplicateDeviceHandler;
+        ConnectionManager.Instance.DuplicateDeviceHandler = null;
+    }
+
+    [TestCleanup]
+    public void TestCleanup()
+    {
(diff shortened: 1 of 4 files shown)


tylerkron:
Superseded by #488 — consolidated the close-race + empty-selection fix onto the visual-redesign branch since they touch the same buttons and code paths. New commit [e6b13b8](https://github.com/daqifi/daqifi-desktop/pull/488/commits/e6b13b8) carries the same change (VM `CloseRequested` event, empty-input guards, idempotent `Close`) plus the unit tests.

PR #552, opened 2026-06-02
fix: skip redundant WiFi version probe during firmware update

tylerkron:
## Summary

`DaqifiViewModel.UpdateWifiModuleAsync` already probes the WiFi firmware version (via `ILanChipInfoProvider`) to decide whether to skip the update and to surface the current/target version in the UI — it **returns early** when the module is already up to date ([DaqifiViewModel.cs:803](Daqifi.Desktop/ViewModels/DaqifiViewModel.cs)). It then called Core's `UpdateWifiModuleAsync`, which — as of `Daqifi.Core 0.22.0` — runs **its own internal version probe by default**, so the device was queried **twice** per WiFi update.

This passes the new `skipVersionCheck: true` parameter so Core flashes directly without re-probing. The 0.22.0 XML docs explicitly recommend this for callers that already did the check (*"Pass `skipVersionCheck: true` to that call to avoid a second probe — see issue #143 for the motivating callsite"*).

## Behavioral note (important for review + on-device testing)

The flag is passed **unconditionally**. The desktop reaches the Core call in four paths, but has only truly "done the version check" in the first:

| Path | Desktop state | Effect of `skipVersionCheck: true` |
|---|---|---|
| chip info present, **update needed** | Flash confirmed | No behavior change |
| chip info **null** (probe failed after retries) | Unsure — "continuing with update" | Was: Core re-probes (may skip). **Now: always flashes** |
| latest-release metadata **null** | Unsure — "continuing" | Same as above |
| device is not `ILanChipInfoProvider` | Never probed | Moot — Core can't probe `coreDevice` either |

In the two "unsure" paths this **removes Core's fallback probe**: a user-initiated update will now re-flash the WiFi module even if it happened to already be current, rather than Core short-circuiting. This is idempotent and matches the desktop's logged intent in those paths, so it is considered acceptable — but it is a real behavioral change, and **there is no unit coverage for the probe-failure paths**.

## ⚠️ Needs hardware validation before merge

I do not have a device attached, so this has **not** been validated against hardware. Per `CLAUDE.md` / [Daqifi.Desktop.UITest/README.md](Daqifi.Desktop.UITest/README.md), run it through the **FlaUI integration gate against a physically connected device**, and specifically exercise:
1. The normal path (out-of-date WiFi module flashes and reports progress correctly).
2. **A probe-failure path** (device that doesn't report chip info, or with release metadata unavailable) to confirm the now-unconditional flash behaves acceptably.

Compile + unit tests pass on CI; that does not cover the on-device flash path.

## Changes

- `DaqifiViewModel.cs`: pass `skipVersionCheck: true`; update the now-stale comment that said Core also probes.
- `DaqifiViewModelFirmwareUpdateTests.cs`: tighten the `UpdateWifiModuleAsync` `Verify` to assert the flag is `true` (the `Setup`/`Callback` stay permissive so they still exercise device routing).

## Stacking

Stacked on **#551** (base branch `deps/nuget-version-alignment`) because `skipVersionCheck` only exists in `Daqifi.Core 0.22.0`, which #551 introduces — it will not compile against current `main`. After #551 squash-merges and its branch is deleted, GitHub auto-retargets this PR to `main`; confirm the diff still shows only the 2 files (rebase onto `main` only if the squash causes divergence).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,8 +1,10 @@
 name: Release
 
 on:
-  release:
-    types: [created]
+  push:
+    tags:
+      - '[0-9]*.[0-9]*.[0-9]*'
+      - 'v[0-9]*.[0-9]*.[0-9]*'
 
 jobs:
   build-and-release:
@@ -30,9 +32,26 @@ jobs:
         cd Daqifi.Desktop.Setup
         dotnet build -c Release
 
-    - name: Upload Release Asset
+    - name: Verify MSI
+      shell: pwsh
+      run: |
+        $msiPath = "Daqifi.Desktop.Setup/DAQifiDesktopSetup/bin/x86/Release/DAQifiDesktop_Setup.msi"
+        if (-not (Test-Path $msiPath)) {
+          Write-Error "MSI not found: $msiPath"
+          exit 1
+        }
+        $size = (Get-Item $msiPath).Length
+        if ($size -eq 0) {
+          Write-Error "MSI is zero bytes: $msiPath"
+          exit 1
+        }
+        Write-Host "MSI OK: $msiPath ($([math]::Round($size / 1MB, 2)) MB)"
+
+    - name: Create Draft Release
       uses: softprops/action-gh-release@v3
       with:
+        draft: true
+        generate_release_notes: true
         files: Daqifi.Desktop.Setup/DAQifiDesktopSetup/bin/x86/Release/DAQifiDesktop_Setup.msi
       env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} 
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -244,6 +244,30 @@ public void InitializeFirewallRules()
 - Squash merge to main
 - All PRs require code review
 
+## Release Process
+
+The release pipeline is compatible with GitHub's **Immutable Releases** setting (which must remain enabled).
+
+**How to cut a release:**
+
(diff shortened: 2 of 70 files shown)


tylerkron:
**Re: Qodo finding 1 — "Always flashes on unknown versions": disagreed, leaving the flag unconditional.** Rationale, checked against Daqifi.Core 0.22.0's shipped XML docs:

1. **Core's internal gate defaults to flashing on unknowns anyway.** The `WifiFirmwareStatus.IsUpToDate` docs state: *"True only when both versions are available AND the device version is at least the latest release. Any unknown is reported as false so the caller defaults to 'needs update'."* In exactly the failure paths this finding worries about (chip probe failed, release metadata unavailable), Core's fallback probe would reach the same "unknown" verdict and proceed to flash — it was never a meaningful version gate there.

2. **Both probes query the same object moments apart.** `SerialStreamingDevice.GetLanChipInfoAsync` delegates to the same underlying core device Core's internal probe would hit. If the desktop's probe failed after its full retry loop, an immediate re-probe of the same device (or a re-query of the same GitHub endpoint on the same network) is overwhelmingly likely to fail the same way. Conditionally re-enabling Core's probe would mostly just add a second slow, doomed probe before a flash that happens regardless.

3. **Unconditional `skipVersionCheck: true` after a caller-side check is Core's documented usage** — the 0.22.0 docs for the parameter cite this very callsite (daqifi/daqifi-core#143) as the motivating example.

The only behavior actually lost is a narrow race (desktop probe fails transiently, Core's immediate re-probe succeeds *and* finds the module current), and its consequence is an idempotent re-flash during an explicitly user-initiated update, on a path where the UI already says "continuing with update."

Also note: the branch has been updated with `main` (post-#551 squash), so the diff now shows only the two real files.

PR #581, opened 2026-06-09
fix(ui): autosize error/success dialogs to content within bounds (#576)

### LINKED ISSUE #576: Error dialogs clipped by fixed 300×200 size — autosize to content (with bounds)
## Summary

The **ERROR** dialog (`Daqifi.Desktop/View/ErrorDialog.xaml`) is a fixed `300 × 200` window with `ResizeMode="NoResize"`. Longer messages overflow / sit cramped against the edges. The firmware **"Waiting for HID bootloader device" (`WaitingForBootloader`)** error is the case that prompted this — the message + "Suggested recovery" text barely fits and crowds the window edges (screenshot attached).

## Repro

Trigger a firmware update failure at the *Waiting for HID bootloader device* stage (device doesn't enter bootloader mode), or any error whose message is longer than a couple of short lines.

## Root cause

`Daqifi.Desktop/View/ErrorDialog.xaml` line 6:

```xml
<controls:MetroWindow ... Title="Error" ResizeMode="NoResize" Width="300" Height="200" ...>
```

Fixed `Width`/`Height` + `NoResize` means the dialog can't grow for longer `ErrorMessage` content, so text wraps into a too-small box and crowds the edges.

## Suggested fix — autosize to content, with bounds

Let error windows size to their content within sane min/max bounds (so short messages stay compact and long ones grow but don't run off-screen). Concretely on the `MetroWindow`:

- `SizeToContent="WidthAndHeight"`
- Replace fixed `Width`/`Height` with **floors** `MinWidth`/`MinHeight` (e.g. `300`/`150`) and **caps** `MaxWidth`/`MaxHeight` (e.g. `560`/`400`) — the cap is what makes long text wrap instead of stretching the window wide.
- Give the message `TextBox` a `MaxWidth` (or rely on the window `MaxWidth`) so `TextWrapping="Wrap"` actually kicks in, and add a little `Margin`/`Padding` so text doesn't touch the edges.

Sketch:

```xml
<controls:MetroWindow ... Title="Error" ResizeMode="NoResize"
        SizeToContent="WidthAndHeight"
        MinWidth="300" MinHeight="150" MaxWidth="560" MaxHeight="400"
        GlowBrush="{DynamicResource AccentColorBrush}">
    <Grid Margin="12">
        ...
        <TextBox Grid.Column="1" Text="{Binding ErrorMessage}" MaxWidth="420"
                 TextWrapping="Wrap" IsReadOnly="True" BorderThickness="0"
                 VerticalAlignment="Center" Foreground="Black" Margin="8,0,0,0"/>
        ...
    </Grid>
</controls:MetroWindow>
```

## Scope

Apply the same autosize-with-bounds pattern to **all** error/alert dialogs for consistency, not just this one — so any future long message renders cleanly rather than being clipped.

## Attachment

`errorwindow.png` — the `WaitingForBootloader` error at the current fixed size (text crowding the edges).
Image

tylerkron:
Fixes #576.

## Problem

The **Error** and **Success** dialogs were a fixed `300×200` `MetroWindow` with `ResizeMode="NoResize"` and the OK button *overlaid* on the icon+message grid. Long messages — the case that prompted the issue is the multi-paragraph **"Waiting for HID bootloader device"** firmware error (a summary line + a "Suggested recovery:" paragraph) — wrapped into the cramped box and clipped against the edges.

The recent dark-mode pass (#580) recolored these dialogs but left the sizing untouched, so the clipping was still live.

## Change

Both dialogs now autosize to their content within bounds:

- `SizeToContent="WidthAndHeight"` with `MinWidth="320" MinHeight="150"` floors and a `MaxWidth="520"` cap → short messages stay compact, long ones wrap (at the message `TextBlock`'s `MaxWidth="400"`) and grow in height instead of stretching wide.
- The message sits in a **height-capped `ScrollViewer`** (`MaxHeight="360"`, `VerticalScrollBarVisibility="Auto"`) so an over-long message scrolls rather than clipping off-screen — the original bug can't recur at any length.
- The **OK button moves into its own grid row** (it was overlaid before, only "working" thanks to the fixed-height slack) and gains `IsDefault="True"` so **Enter** dismisses the dialog.
- Dark-mode tokens from #580 (`Surface`, `TextPrimary`, `StatusRed`/`StatusGreen`) are preserved, and the layout now matches `DuplicateDeviceDialog`'s conventions (20px margins, 40px icon, 80×30 buttons) so the alert family reads as one system.

## Verification

Built clean, and rendered the real windows against a long and a short message (measured via `GetWindowRect`):

| Message | Window size |
|---|---|
| Firmware error (summary + recovery paragraph) | **492 × 232** — wraps, fully visible, no clipping |
| "Device disconnected unexpectedly." | **320 × 160** — compact, at the min-bound |

Both were a fixed `300 × 200` before. The long message is the exact `WaitingForBootloader` text from the issue; it now renders in full with the OK button in its own row.

## Scope note

`DuplicateDeviceDialog` was intentionally left unchanged: it was already redesigned in #580 with a proper row layout, and its only variable text (device serial + interface string) is length-bounded, so it doesn't clip. Happy to extend the same autosize-with-bounds pattern there if reviewers prefer strict "all dialogs" consistency.

### DIFF
--- a/Daqifi.Desktop/View/DuplicateDeviceDialog.xaml
+++ b/Daqifi.Desktop/View/DuplicateDeviceDialog.xaml
@@ -3,7 +3,10 @@
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
         xmlns:controls="clr-namespace:MahApps.Metro.Controls;assembly=MahApps.Metro"
         xmlns:iconPacks="http://metro.mahapps.com/winfx/xaml/iconpacks"
-        Title="Device Already Connected" ResizeMode="NoResize" Width="480" Height="280" BorderThickness="0" Background="{StaticResource Surface}" GlowBrush="{DynamicResource AccentColorBrush}">
+        Title="Device Already Connected" ResizeMode="NoResize"
+        SizeToContent="WidthAndHeight"
+        MinWidth="480" MinHeight="280" MaxWidth="620" MaxHeight="520"
+        BorderThickness="0" Background="{StaticResource Surface}" GlowBrush="{DynamicResource AccentColorBrush}">
 
     <controls:MetroWindow.Resources>
         <ResourceDictionary>
@@ -13,6 +16,8 @@
         </ResourceDictionary>
     </controls:MetroWindow.Resources>
 
+    <!-- Autosizes to content within bounds (DAQiFi#576); MinWidth/MinHeight hold the prior
+         480x280 footprint for the common case, so only unusually long device/interface text grows it. -->
     <Grid Margin="20" Background="{StaticResource Surface}" TextElement.Foreground="{StaticResource TextPrimary}">
         <Grid.RowDefinitions>
             <RowDefinition Height="Auto"/>
@@ -33,7 +38,7 @@
 
         <!-- Message -->
         <TextBlock Grid.Row="0" Grid.Column="1" Text="{Binding Message}"
-                  VerticalAlignment="Top" Margin="10,5,0,10"
+                  VerticalAlignment="Top" Margin="10,5,0,10" MaxWidth="385"
                   TextWrapping="Wrap" FontSize="14" Foreground="{StaticResource TextPrimary}"/>
 
         <!-- Options (sit directly under the prompt, aligned with the message) -->
--- a/Daqifi.Desktop/View/ErrorDialog.xaml
+++ b/Daqifi.Desktop/View/ErrorDialog.xaml
@@ -10,9 +10,10 @@
         GlowBrush="{DynamicResource AccentColorBrush}">
     <!--
         Autosizes to the message within bounds (DAQiFi#576): MinWidth/MinHeight keep short
-        messages compact, MaxWidth + the TextBlock's MaxWidth force long text to wrap rather
-        than stretch the window wide, and the height-capped ScrollViewer guarantees an
-        over-long message scrolls instead of clipping off-screen.
+        messages compact, the read-only TextBox's MaxWidth forces long text to wrap rather
+        than stretch the window wide, and its MaxHeight + Auto vertical scrollbar guarantee an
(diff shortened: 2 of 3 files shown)


tylerkron:
Thanks @qodo-code-review — both findings addressed in b8c6f16.

**1. `DuplicateDeviceDialog` still fixed-size (Requirement gap) — Fixed.**
Applied the autosize-with-bounds pattern: `SizeToContent="WidthAndHeight"` with `MinWidth="480" MinHeight="280"` (the prior fixed footprint, so the common case is visually unchanged) and `MaxWidth="620" MaxHeight="520"` caps, plus `MaxWidth` on the message `TextBlock` so long text wraps. Verified by rendering with a long interface string (`WiFi (192.168.100.123:9760)`): the dialog measures **483×280** and the message + both radio options render without clipping, buttons anchored at the bottom.

**2. Dialog text not selectable (Correctness) — Fixed.**
Reverted the Error/Success message from `TextBlock` back to a read-only (not disabled) `TextBox` — `IsReadOnly="True"`, chrome-less (`BorderThickness="0" Background="Transparent"`), `TextWrapping="Wrap"`, `MaxWidth="400" MaxHeight="360"`, `VerticalScrollBarVisibility="Auto"` — matching the suggested approach. This restores select/copy for the multi-paragraph firmware error + recovery text. I'd been concerned a wrapping `TextBox` under `SizeToContent` might greedily widen short messages, but verified it does not: the short error stays compact at **322×160** while the long firmware error grows to **492×240**, with the text selectable.

On the summary's alternative approaches (shared `AlertDialog` template / `DialogCoordinator`): reasonable, but leaving them as a possible future follow-up — for two near-identical dialogs the abstraction isn't yet warranted and it's out of scope for this bug fix.

PR #585, opened 2026-06-10
fix(ci): make release workflow compatible with immutable releases

### LINKED ISSUE #569: fix(ci): make release workflow compatible with immutable releases
## Context

Cutting **3.2.0** surfaced that our release pipeline is incompatible with GitHub's **Immutable Releases** feature. To ship 3.2.0 with its installer we had to *disable* immutable releases on the repo. We want to **re-enable** immutable releases (artifact tamper-protection / supply-chain integrity) without breaking the MSI upload.

## What happens today

`.github/workflows/release.yaml` triggers on `release: [created]`. The flow is:

1. A maintainer creates the GitHub Release (**published**) with curated notes.
2. CI builds the MSI and uploads it to the **already-published** release via `softprops/action-gh-release@v3`.

With immutable releases **ON**, step 2 is rejected:

```
Cannot upload asset DAQifiDesktop_Setup.msi to an immutable release.
GitHub only allows asset uploads before a release is published,
so upload assets to a draft release before you publish it.
```

Net effect: the release publishes with notes but **no installer**. Worse — once a release is published while immutable, that release object is **permanently immutable**: it cannot be reverted to draft or have assets added later, and must be deleted and recreated. (We hit exactly this on 3.2.0: HTTP 422 `state cannot be changed when release is immutable`.)

## Root cause

The pipeline attaches the artifact **after** the release is published. Immutable releases require every asset to be present **before** publish — i.e. attach to a **draft**, then publish.

## Goal

Re-enable the repo's "Immutable releases" setting and have the pipeline attach `DAQifiDesktop_Setup.msi` reliably, while preserving human-curated release notes.

## Proposed fix — attach-while-draft, then publish

**Option A (recommended) — keep curated manual notes:**
- Trigger on tag push: `on: push: tags: ['[0-9]+.[0-9]+.[0-9]+']`.
- CI builds the MSI and creates a **draft** release for that tag with the MSI attached (`softprops/action-gh-release@v3` with `draft: true`, `files:`, `generate_release_notes: true` as a starting point).
- A maintainer reviews/edits the draft notes and clicks **Publish**. Because the asset was added while draft, publishing is immutable-safe.

**Option B — fully automated:**
- Same tag-push trigger; CI builds, attaches the MSI to a draft, then flips it to published in a follow-up step (`gh release edit <tag> --draft=false`). Notes are auto-generated (less curation).

Either way:
- Guard the upload so the job fails loudly if the MSI is missing or zero bytes.
- Consider pinning the Windows runner (`windows-latest` redirects to `windows-2025-vs2026` by 2026-06-15).

## Acceptance criteria

- [ ] Repo **Immutable releases** setting is re-enabled.
- [ ] Cutting a new version produces a **published** release with `DAQifiDesktop_Setup.msi` attached.
- [ ] The flow preserves (or allows) human-curated release notes.
- [ ] Release process documented in `CLAUDE.md` / README.

## References

- Failed run (immutable upload rejection): https://github.com/daqifi/daqifi-desktop/actions/runs/27040611713
- Workflow file: `.github/workflows/release.yaml`
- softprops/action-gh-release immutable-release guidance.

tylerkron:
## Summary

- Switches trigger from `release: [created]` to `push: tags: v[0-9]+.[0-9]+.[0-9]+` so CI runs on a tag push instead of after a release is already published
- Adds a "Verify MSI" step that fails loudly if the installer file is missing or zero bytes
- Creates a **draft** release with the MSI attached (via `softprops/action-gh-release@v3 draft: true`), then a maintainer edits the auto-generated notes and clicks Publish — immutable-safe because the asset was added while still draft
- Documents the new release flow in `CLAUDE.md`

> **Manual step required after merge**: re-enable the repo's "Immutable releases" setting under Settings → General → Releases. It was disabled to ship 3.2.0 and should be turned back on once this PR lands.

## Why this fixes it

GitHub's immutable releases feature rejects asset uploads to an already-published release (HTTP 422). The old pipeline attached the MSI *after* publication, which was always incompatible. The new pipeline attaches assets to a draft before the human publishes it, which is exactly what the immutable-releases model expects.

## Test plan

- [ ] Push a test tag (`v0.0.0-test`) to a fork and confirm the workflow runs, builds the MSI, and creates a **draft** release with the installer attached
- [ ] Verify the draft can be edited and published without error
- [ ] Re-enable the repo's Immutable releases setting after merging
- [ ] Cut the next real release (e.g. `v3.3.0`) and confirm the end-to-end flow

Closes #569


tylerkron:
Thanks for the review, @qodo-code-review!

**Pre-release tag pattern** — noted for the future. The current regex (`v[0-9]+.[0-9]+.[0-9]+`) is intentionally strict since the project has no pre-release workflow today. We can loosen it to `v*` or add an `-rc` variant when the need arises.

No other action items from this review.

tylerkron:
Addressing Qodo's review findings:

**1. Tag glob matches wrong (Action required)** — Fixed. Changed `v[0-9]+.[0-9]+.[0-9]+` to `v[0-9]*.[0-9]*.[0-9]*`. Correct — GitHub Actions tag filters use glob matching, not regex, so `+` was being treated as a literal character rather than "one or more digits". `v3.3.0` would never have triggered the workflow as written.

**2. CLAUDE.md lines exceed 120 chars (Review recommended)** — Fixed. Wrapped the long lines in the Release Process section to stay within 120 characters.

**3. Release docs out of sync (Optional)** — Fixed. Updated README.md "For maintainers" section to describe the new tag-push → draft → publish flow, removing the stale `release: created` description.

Issue #592, opened 2026-06-12
Break up the two god classes: DaqifiViewModel (2,229 lines) and DatabaseLogger (1,820 lines)


tylerkron:
## Context

The two largest classes in the app each mix several unrelated responsibilities, which makes them hard to navigate, hard to unit-test, and the most expensive files for both humans and AI agents to work in. This issue proposes behavior-preserving extractions, **one PR per extraction** — do not attempt this as a single PR. Verified against `main` @ 287c005.

Related: #591 covers the device layer; no overlap with this issue.

## 1. `Daqifi.Desktop/ViewModels/DaqifiViewModel.cs` (2,229 lines)

Verified responsibility map:

| Cluster | Lines (approx.) | Contents |
|---|---|---|
| Firmware update orchestration | 674–1129 (`#region Updload firmware…` — note region name typo) + 1712–1843 (version checking) | `UploadFirmware`, cancel logic, `CreateWifiFirmwareUpdateService`, `NormalizeWifiFirmwareVersion`, `HandleFirmwareUpdateException`, `ShowFirmwareErrorDialog`, service factories, `GetFirmwareupdatationList` (typo'd name — fix while moving), update notifications |
| Logging-session list actions | 1301–1663 | `DisplayLoggingSession` (uses legacy `BackgroundWorker`), `ExportLoggingSession`, `ExportAllLoggingSession`, `DeleteLoggingSessionAsync`, `DeleteAllLoggingSessionAsync`, collection-changed plumbing |
| Network settings status | 1187–1257 | `UpdateNetworkConfiguration`, status reset/auto-clear, CTS management |
| Confirm overlay + message dialogs | 2047–2095 | `ShowMessage`, `ShowConfirm`, `CompleteConfirm` |
| Disk space monitoring | 2166–2229 | own `#region` |
| `async void UpdateUi` | 1959 | PropertyChanged handler (do not fix here — see out-of-scope) |

### Proposed extractions (in order of value)

[...]

PR #603, opened 2026-06-16
chore(viewmodels): extract ConfirmOverlayViewModel from DaqifiViewModel (#592)

tylerkron:
## Summary

Third "small extraction" from #592 (after `FirmwareUpdateCoordinator` #601 and `LoggingSessionListViewModel` #602). Extracts the in-pane confirm-overlay cluster out of `DaqifiViewModel` into a new reusable, **WPF-free, unit-testable** `ConfirmOverlayViewModel : ObservableObject`.

The new class owns the bound state (`IsOpen`, `Title`, `Message`, `AffirmativeLabel`, `AffirmativeIsDestructive`), the affirmative/negative commands, and an awaitable `ShowAsync(...)` built on the `TaskCompletionSource` pattern (`RunContinuationsAsynchronously`). No `Application.Current`, dispatcher, or singletons — the result resolves purely when the two commands fire.

## What changed

- **`DaqifiViewModel`** exposes the overlay as a nested `public ConfirmOverlayViewModel ConfirmOverlay { get; }` and keeps `ShowConfirm` as a **thin forwarder**, so the existing `ILoggingSessionListHost.ShowConfirmAsync` caller is unchanged. The pane-navigation reset path (`SelectedIndex` setter, formerly `CompleteConfirm(false)`) now calls `ConfirmOverlay.Cancel()`.
- **`ProfilesPaneViewModel`** *also* adopts the same `ConfirmOverlayViewModel`. ⚠️ Note: the Profiles pane binds **its own** `ProfilesPaneViewModel` (set in `ProfilesPane.xaml.cs` on `Loaded`), **not** `DaqifiViewModel` — and it carried a complete **duplicate** confirm-overlay implementation. This PR deletes that duplicate and routes it through the shared component, which is exactly the cross-pane reuse #592 calls for. Its `Cleanup()` reset path now calls `ConfirmOverlay.Cancel()`.
- **XAML** — both confirm-overlay blocks re-pointed to bind `ConfirmOverlay.*`:
  - `Daqifi.Desktop/View/Prototype/LoggedDataPanePrototype.xaml`
  - `Daqifi.Desktop/View/ProfilesPane.xaml`
  - In `LoggedDataPanePrototype.xaml` the overlay `Grid`'s `x:Name` was renamed `ConfirmOverlay` → `ConfirmOverlayHost` to avoid colliding with the new DataContext property (the generated field was unreferenced in code-behind; `ConfirmCancelButton` and `OnConfirmOverlayIsVisibleChanged` are unchanged).
- **Docs** — `Daqifi.Desktop.UITest` comment/README references updated to the new binding names. The `ConfirmAffirmativeButton` AutomationId the FlaUI harness drives is unchanged.

## Behavior preservation

Logic is preserved verbatim: the re-entrancy guard (a still-pending confirm resolves `false` before a new one opens), `TaskCreationOptions.RunContinuationsAsynchronously`, and the open/close sequencing all carry over unchanged. `ShowMessage` / `ShowDiskSpaceMessage` (MahApps `MetroWindow.ShowMessageAsync` modals) are intentionally **left in place** — a separate mechanism.

`DaqifiViewModel` shrinks further toward the issue's ~800-line target.

## Tests

New `Daqifi.Desktop.Test/ViewModels/ConfirmOverlayViewModelTests.cs` drives the VM with no WPF:
`ShowAsync` opens with the right title/message/label/destructive flag (+ `IsOpen`); affirmative resolves `true` & closes; negative resolves `false` & closes; a second `ShowAsync` while one is pending resolves the first `false`; `Cancel()` reset path; `RunContinuationsAsynchronously` (continuation not run inline); destructive-flag/label reset on non-destructive re-open; and double-resolve idempotency.

## Verification

- `dotnet build "DAQiFi Desktop.sln"` — **0 errors**.
- Unit gate `dotnet test --filter "TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests"` — **all pass** (incl. the 12 new `ConfirmOverlayViewModel` tests; existing `DaqifiViewModel` / `LoggingSessionListViewModel` tests stay green).
- Multi-agent adversarial review (behavior / XAML-binding / tests / standards dimensions) — clean on code; the two test-coverage findings it surfaced are included here.
- The freshly-built app launches and loads both panes' XAML without error.
- ⚠️ The interactive click-through of both live overlays could not be completed in this environment (the computer-use screen-access approval dialog timed out). Binding correctness is otherwise covered by the exact property-name match, the adversarial binding review, and the clean app launch; a manual click-through (delete-confirm in Logged Data, switch-confirm in Profiles) is recommended before merge.

### DIFF
--- a/Daqifi.Desktop.Test/ViewModels/ConfirmOverlayViewModelTests.cs
+++ b/Daqifi.Desktop.Test/ViewModels/ConfirmOverlayViewModelTests.cs
@@ -1,0 +1,225 @@
+using Daqifi.Desktop.ViewModels;
+
+namespace Daqifi.Desktop.Test.ViewModels;
+
+/// <summary>
+/// Behavior contract for <see cref="ConfirmOverlayViewModel"/> — the reusable in-pane confirm
+/// overlay extracted from <c>DaqifiViewModel</c> (issue #592). The overlay is driven with no WPF
+/// dependency: await <see cref="ConfirmOverlayViewModel.ShowAsync"/> and fire the affirmative /
+/// negative commands (or <see cref="ConfirmOverlayViewModel.Cancel"/>). Covers the opening state,
+/// the two command resolutions, the re-entrancy guard, the host reset path, and the
+/// RunContinuationsAsynchronously contract.
+/// </summary>
+[TestClass]
+public class ConfirmOverlayViewModelTests
+{
+    #region Initial state / ShowAsync
+
+    [TestMethod]
+    public void NewOverlay_IsClosedWithDefaults()
+    {
+        var overlay = new ConfirmOverlayViewModel();
+
+        Assert.IsFalse(overlay.IsOpen);
+        Assert.AreEqual(string.Empty, overlay.Title);
+        Assert.AreEqual(string.Empty, overlay.Message);
+        Assert.AreEqual("OK", overlay.AffirmativeLabel);
+        Assert.IsFalse(overlay.AffirmativeIsDestructive);
+    }
+
+    [TestMethod]
+    public void ShowAsync_OpensOverlayWithSuppliedContent()
+    {
+        var overlay = new ConfirmOverlayViewModel();
+
+        var task = overlay.ShowAsync(
+            "Delete Confirmation",
+            "Are you sure you want to delete Bench Run?",
+            affirmativeLabel: "DELETE",
+            isDestructive: true);
+
+        Assert.IsTrue(overlay.IsOpen);
+        Assert.AreEqual("Delete Confirmation", overlay.Title);
+        Assert.AreEqual("Are you sure you want to delete Bench Run?", overlay.Message);
+        Assert.AreEqual("DELETE", overlay.AffirmativeLabel);
+        Assert.IsTrue(overlay.AffirmativeIsDestructive);
+        Assert.IsFalse(task.IsCompleted, "The task stays pending until the user responds.");
+    }
+
+    [TestMethod]
+    public void ShowAsync_DefaultsToNonDestructiveOkLabel()
+    {
+        var overlay = new ConfirmOverlayViewModel();
+
+        overlay.ShowAsync("Switch profile?", "Switch to 'Bench'?");
+
+        Assert.AreEqual("OK", overlay.AffirmativeLabel);
+        Assert.IsFalse(overlay.AffirmativeIsDestructive);
+        Assert.IsTrue(overlay.IsOpen);
+    }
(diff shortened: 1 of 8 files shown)


tylerkron:
Thanks @qodo-code-review — addressing the two rule violations from the summary:

**1. Long `Visibility` binding — `LoggedDataPanePrototype.xaml` (📘 fixed)**

Wrapped the affirmative-button `Visibility` markup extension onto a continuation line so it's under 120 chars, and wrapped the symmetric danger-style variant (the other affirmative button, previously 134 chars) for consistency. Valid XAML, no behavior change — verified the project still builds. Fixed in bf80135.

**2. Long README line — `Daqifi.Desktop.UITest/README.md:363` (📘 respectfully disagree — false positive)**

That line measures **109 characters**, which is within the 120-char limit, so there's nothing to change:

```
    `Visibility` binds to `ConfirmOverlay.AffirmativeIsDestructive`, and a `Collapsed` subtree is absent from
```

(`awk '{print length}'` → 109. Happy to reflow it if your rule is counting differently.)

Also — thanks for the 🐞 0-bug result and the High-Level Assessment endorsing the small, WPF-free `ConfirmOverlayViewModel` extraction as the best fit here.

tylerkron:
**Update — live visual verification completed** (computer-use access worked after a Claude restart).

**Logged Data pane — verified end-to-end against the re-pointed `ConfirmOverlay.*` bindings.** Triggered the per-row delete on a real session: the in-pane confirm overlay rendered with title *"Delete Confirmation"*, message *"Are you sure you want to delete Session_0?"*, the red/danger **DELETE** affirmative button, and **CANCEL**. So `ConfirmOverlay.IsOpen` / `Title` / `Message` / `AffirmativeLabel` / `AffirmativeIsDestructive` all resolve. Clicking **CANCEL** closed the overlay and left the session intact — `ConfirmOverlay.NegativeCommand` fires, no DB write.

**Profiles pane** renders cleanly with its re-pointed bindings (its `ProfilesPaneViewModel` now exposes the same `ConfirmOverlay`). Its overlay uses the identical `ConfirmOverlay.*` shape and the same component verified above + unit-tested; the only path that opens it (`ActivateProfile` switching between two active-matching profiles) requires a connected device and two matching profiles, which I didn't stage — so it's confirmed structurally (clean render + identical binding) rather than via a full live switch.

Net: the binding refactor is confirmed working in the running app, in addition to the green CI build/test and the clean adversarial review.

PR #609, opened 2026-06-18
chore(viewmodels): extract DiskSpaceMonitorCoordinator from DaqifiViewModel (#592)

tylerkron:
## What

Extracts the disk-space gating + monitoring logic out of `DaqifiViewModel` into a WPF-free, unit-testable `DiskSpaceMonitorCoordinator` — one of the small VM-side extractions in #592.

**New (`Daqifi.Desktop/DiskSpace/`):**
- `DiskSpaceMonitorCoordinator` — owns the pre-logging gate (`EvaluateStartLogging`: block on critical, warn on low-but-not-critical), the in-session monitor start/stop, and the low/critical threshold event handling (warn, or auto-stop logging). Constructor-injected with `IDiskSpaceMonitor` + `IAppLogger` — no `AppLogger.Instance` / `App.*` reach-ins.
- `IDiskSpaceMonitorHost` — the narrow seam the coordinator uses to reach back for the two WPF concerns it can't own (stop the session, present a dialog), mirroring the existing `IFirmwareUpdateHost` / `ILoggingSessionListHost` precedent.
- `DiskSpaceStartDecision` — the gate verdict.

`DaqifiViewModel` keeps the host-seam implementation (which marshals those two calls to the UI thread — the monitor's threshold events fire on its background timer thread) and the MahApps dialog helper.

## Behavior-preserving

All dialog text, the suppress-initial-warning plumbing, the TwoWay-toggle revert on a blocked start, the critical → stop-logging → dialog ordering, and the non-blocking `BeginInvoke` marshalling are unchanged. The composition root still builds the monitor only during window init, so non-window-init construction paths leave the coordinator null and skip the gate exactly as before.

## Honest note on line count

`DaqifiViewModel` stays at **1,607 lines (net zero)**: the extracted logic is replaced, roughly line-for-line, by the UI-thread-marshalling host seam that must remain in the view model. **The win here is cohesion + test coverage** (the gate/event logic previously had none because it was WPF-coupled), **not line reduction**. This is consistent with the discussion in #592 that `DaqifiViewModel`'s ~800-line target is unreachable by behavior-preserving extraction; the only remaining extraction that meaningfully reduces line count is the DatabaseLogger `ViewportController`.

## Tests

Adds `DiskSpaceMonitorCoordinatorTests` (16 tests): gate verdicts, exact dialog wording (pre-logging vs in-session warning differ), suppress-warning flag, auto-stop-before-dialog ordering, event unsubscription on dispose, and constructor guards.

Verification: build clean (0 errors); full unit gate **450 passed / 0 failed** (`dotnet test --filter "TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests"`).

Relates to #592.

### DIFF
--- a/Daqifi.Desktop.Test/DiskSpace/DiskSpaceMonitorCoordinatorTests.cs
+++ b/Daqifi.Desktop.Test/DiskSpace/DiskSpaceMonitorCoordinatorTests.cs
@@ -1,0 +1,282 @@
+using Daqifi.Desktop.Common.Loggers;
+using Daqifi.Desktop.DiskSpace;
+using Moq;
+
+namespace Daqifi.Desktop.Test.DiskSpace;
+
+/// <summary>
+/// Behavior contract for <see cref="DiskSpaceMonitorCoordinator"/>. The disk-space gate, the
+/// in-session monitor delegation, and the low/critical event handling were extracted from
+/// <c>DaqifiViewModel</c> (issue #592); these tests target the coordinator directly through a
+/// lightweight <see cref="FakeDiskSpaceMonitorHost"/> and <see cref="FakeDiskSpaceMonitor"/>, with no
+/// WPF dependency. The asserted behavior (block/warn verdicts, dialog text, auto-stop on critical,
+/// suppress-initial-warning plumbing) is unchanged from the pre-refactor view model.
+/// </summary>
+[TestClass]
+public class DiskSpaceMonitorCoordinatorTests
+{
+    private const long MB = 1024 * 1024;
+
+    private FakeDiskSpaceMonitorHost _host = null!;
+    private FakeDiskSpaceMonitor _monitor = null!;
+    private DiskSpaceMonitorCoordinator _coordinator = null!;
+
+    [TestInitialize]
+    public void Setup()
+    {
+        _host = new FakeDiskSpaceMonitorHost();
+        _monitor = new FakeDiskSpaceMonitor();
+        _coordinator = new DiskSpaceMonitorCoordinator(_host, _monitor, Mock.Of<IAppLogger>());
+    }
+
+    #region EvaluateStartLogging Tests
+
+    [TestMethod]
+    public void EvaluateStartLogging_OkSpace_ReturnsAllowed_AndShowsNoDialog()
+    {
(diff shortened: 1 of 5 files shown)


tylerkron:
Thanks @qodo-code-review — responses to the three findings:

**🐞 Off-thread stop logging (Bug) — fixed** in 84e2bdc. Good catch: `IDiskSpaceMonitorHost.StopLogging` was running `IsLogging = false` inline when `Application.Current?.Dispatcher` is null, which could execute the setter (PropertyChanged + `ConnectedDevices` iteration) on the monitor's background timer thread. It now no-ops when there's no dispatcher to marshal to, restoring the original handler's `BeginInvoke` semantics (which silently skipped when the dispatcher was unavailable) and removing the cross-thread risk. The on-thread and off-thread (`BeginInvoke`) paths are unchanged.

**📘 DaqifiViewModel lacks IDisposable (Rule violation) — declined for this PR.** This is a pre-existing pattern, not introduced here: before this extraction the VM already owned an `IDisposable` (`IDiskSpaceMonitor`) and cleaned it up via the same custom `DisposeDiskSpaceMonitor()` method (called from `MainWindow.xaml.cs` on close). This behavior-preserving extraction just swaps the disposed object. Converting the VM to `IDisposable` is a broader lifecycle change — it owns other resources that aren't disposed today (`_networkSettingsAppliedCts`, the `_sdLoggingElapsedTimer` `DispatcherTimer`) and it would change the `MainWindow.xaml.cs` call site — so it belongs in its own change rather than being bundled into this disk-space extraction. Noting it as a good follow-up for the #592 effort.

**📘 DaqifiViewModel instantiates DiskSpaceMonitor (Rule violation) — declined.** By design: `DaqifiViewModel` is the composition root, mirroring the merged `FirmwareUpdateCoordinator` extraction, which likewise builds its production service defaults in the VM (`CreateDefaultFirmwareDownloadService` etc.) and injects them into the coordinator. The coordinator itself is fully DI-friendly — it takes `IDiskSpaceMonitor` and `IAppLogger` as constructor parameters and is unit-tested against fakes — so only the production default is `new`'d at the root. This matches the precedent where injecting `PlotModelFactory`/`CsvExporter` was rejected (#608, #522), which the finding itself notes.

PR #610, opened 2026-06-19
chore(viewmodels): make DaqifiViewModel implement IDisposable (#592)

tylerkron:
## What

Makes `DaqifiViewModel` implement `IDisposable`, consolidating its ad-hoc teardown into a single `Dispose()`. Follow-up to #609 — Qodo flagged the VM owning an `IDisposable` (`DiskSpaceMonitorCoordinator`) without implementing `IDisposable` (rule 244824, citing the accepted `VersionNotification` precedent in #519); it was deferred out of #609 to keep that extraction behavior-preserving.

> **Stacked on #609.** Base is `claude/kind-curie-b9e5ac` so the diff shows only this change; GitHub will auto-retarget to `main` once #609 merges. Review/merge #609 first.

## Change

- Replaces the bespoke public `DisposeDiskSpaceMonitor()` with `IDisposable.Dispose()`; `MainWindow`'s `Closing` handler now calls `Dispose()` (single disposal entry point, same position relative to `ShutdownCommand`).
- `Dispose()` now releases everything the VM owns, in one place:
  - the `DiskSpaceMonitorCoordinator` (all `DisposeDiskSpaceMonitor` did),
  - the transient network-settings status `CancellationTokenSource` (via the existing `CancelAndDisposeNetworkSettingsCts`),
  - the SD-card elapsed-time `DispatcherTimer`,
  - the long-lived `ConnectionManager`/`LoggingManager` `PropertyChanged` handlers + the per-device logging-state handlers, and detaches the session-list VM from the singleton `LoggingManager.LoggingSessions` collection — all of which otherwise pin the VM (and its object graph) for the process lifetime.
- Adds `LoggingSessionListViewModel.DetachCollection()` (mirror of `AttachCollection`) so the list VM can release its singleton-collection subscription.

A `_disposed` guard makes `Dispose()` idempotent; `-=` on never-wired handlers is a no-op, so non-window-init paths (tests) dispose cleanly.

## Behavior-preserving

Disposal still happens once, at window close, in the same order relative to `ShutdownCommand`. The added cleanup only releases resources/subscriptions at shutdown — no runtime behavior changes. `Shutdown()` still disconnects every device (`Dispose()` doesn't clear `ConnectedDevices`, only unsubscribes its change handler).

## Tests

Adds two `DetachCollection` unit tests (unsubscribes the observed collection; safe no-op when nothing was attached). Build clean; full unit gate **452 passed / 0 failed** (`dotnet test --filter "TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests"`).

Relates to #592. Addresses a Qodo finding on #609.

### DIFF
--- a/Daqifi.Desktop.Test/ViewModels/LoggingSessionListViewModelTests.cs
+++ b/Daqifi.Desktop.Test/ViewModels/LoggingSessionListViewModelTests.cs
@@ -424,13 +424,19 @@ public class LoggingSessionListViewModelTests
     }
 
     [TestMethod]
-    public void DetachCollection_WhenNothingAttached_DoesNotThrow()
+    public void DetachCollection_WhenNothingAttached_IsSafeNoOp()
     {
         var host = new FakeLoggingSessionListHost();
         var list = CreateList(host);
 
         // Mirrors the non-window-init path where AttachCollection was never called.
         list.DetachCollection();
+
+        // The no-op detach must leave the list usable: a later attach still wires notifications.
+        var collection = new ObservableCollection<LoggingSession>();
+        list.AttachCollection(collection);
+        collection.Add(new LoggingSession { ID = 1, Name = "A" });
+        Assert.AreEqual(1, host.NotifyCount);
     }
 
     #endregion
--- a/Daqifi.Desktop.Test/ViewModels/LoggingSessionListViewModelTests.cs
+++ b/Daqifi.Desktop.Test/ViewModels/LoggingSessionListViewModelTests.cs
@@ -407,6 +407,32 @@ public class LoggingSessionListViewModelTests
             "Re-attaching the same collection must not add a second subscription.");
     }
 
+    [TestMethod]
+    public void DetachCollection_UnsubscribesFromTheObservedCollection()
+    {
+        var host = new FakeLoggingSessionListHost();
+        var list = CreateList(host);
+        var collection = new ObservableCollection<LoggingSession>();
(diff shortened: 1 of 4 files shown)


tylerkron:
Thanks @qodo-code-review — both addressed in 385c55d:

**🐞 Debug event still subscribed (Bug) — fixed.** Good catch. `Dispose()` now also unsubscribes `DebugDataReceived` for each `AbstractStreamingDevice` in `ConnectedDevices` (the per-device handler wired up in `UpdateConnectedDeviceUI`), so a connected device no longer keeps the disposed VM rooted and a late debug callback can't marshal onto the UI thread during close.

One scoping note: the suggested follow-up of also unsubscribing `DebugDataReceived` when a device is *removed mid-session* (handler accumulation across reconnects, suggestion #2) is a separate pre-existing issue — it predates this PR and isn't part of the IDisposable consolidation. Worth a small dedicated fix (mirroring the `_loggingStateSubscribedDevices` tracking-set pattern for debug subscriptions); happy to file it under #592 if you'd like.

**📘 DetachCollection test lacks Assert (Rule violation) — fixed.** Renamed to `DetachCollection_WhenNothingAttached_IsSafeNoOp` and it now asserts real behavior: after a no-op detach, a subsequent `AttachCollection` + mutation still raises the notification (proving the no-op left the list usable), rather than only relying on "doesn't throw."

Full unit gate green (452 passed).

PR #633, opened 2026-06-23
deps: bump Daqifi.Core to 0.24.0

tylerkron:
## Summary

Bumps **Daqifi.Core 0.23.0 → 0.24.0** across `Daqifi.Desktop.DataModel`, `Daqifi.Desktop.IO`, and `Daqifi.Desktop`.

The upgrade is **not drop-in source-compatible** — it required three adaptations:

| Change | Why |
|---|---|
| `System.Management` 10.0.8 → 10.0.9 (`Daqifi.Desktop.csproj`) | Core 0.24.0 raised its transitive floor to `>= 10.0.9`; the old pin produced an `NU1605` downgrade-as-error at restore. Now consistent with the other `System.*` 10.0.9 packages. |
| `BootloaderSessionStreamingDeviceAdapter` implements 8 new members | Core 0.24.0 (#250, device-level channel management) added `EnableChannel`/`EnableChannels`/`DisableChannel`/`DisableAllChannels`/`SetDioDirection`/`SetDioValue`/`SetAnalogOutput`/`Reboot` to `Daqifi.Core.Device.IStreamingDevice`. The adapter implements the Core interface by hand, so it must provide them — added as **no-ops**, matching its documented bootloader-only "intentionally no-op" design (channel management / outputs / reboot are meaningless mid-firmware-update). The real devices inherit Core's concrete base and were unaffected. |
| `CoreConnectionTemplateTests.TemplateCoreDevice` override signature | Core 0.24.0 (#247) changed `InitializeAsync()` → `InitializeAsync(TimeSpan? channelPopulationTimeout = null, CancellationToken cancellationToken = default)`. Production call sites survive on the defaults; only the test fixture's `override` needed updating. |

## Notable Core 0.24.0 behavior changes (no code change needed, but worth knowing)

- **#247** — `InitializeAsync` now blocks until the device reports its channels (re-sending `GetDeviceInfo` periodically) instead of a fixed delay, and surfaces a `TimeoutException` if channels never populate. More reliable connects; a new failure mode on connect.
- **#253 (breaking, wire-protocol)** — SD-card logging now sends `SYST:STOR:SD:FILE` instead of the pre-3.5.0 `SYST:STOR:SD:LOGging`, so **SD logging requires device firmware v3.5.0+**. No .NET API change.

## Validation (Windows, against an attached Nq1 device)

- `dotnet build "DAQiFi Desktop.sln"` — **clean, 0 errors**.
- Unit suite (fast gate filter) — **454 / 454 pass**.
- FlaUI UI integration gate — **16 / 17 pass**.
  - The single failure (`SdCardLogging_LogsToSdCard_ThenImportsToSession`) is a **pre-existing device-firmware SD-read-over-USB defect, independent of this upgrade**: on a freshly-formatted card the device writes valid log data and enumerates it correctly, but downloading the file over USB returns 0 bytes. Reproduces on pre-upgrade Core; the SD-read path is unchanged since 0.22.0. Tracked separately for firmware/Core investigation.

### DIFF
--- a/Daqifi.Desktop.DataModel/Daqifi.Desktop.DataModel.csproj
+++ b/Daqifi.Desktop.DataModel/Daqifi.Desktop.DataModel.csproj
@@ -8,7 +8,7 @@
 	</PropertyGroup>
 	<ItemGroup>
 	  <PackageReference Include="CommunityToolkit.Mvvm" Version="8.4.2" />
-	  <PackageReference Include="Daqifi.Core" Version="0.23.0" />
+	  <PackageReference Include="Daqifi.Core" Version="0.24.0" />
 	  <PackageReference Include="Roslynator.Analyzers" Version="4.15.0">
 		<PrivateAssets>all</PrivateAssets>
 		<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
--- a/Daqifi.Desktop.IO/Daqifi.Desktop.IO.csproj
+++ b/Daqifi.Desktop.IO/Daqifi.Desktop.IO.csproj
@@ -7,7 +7,7 @@
 	</PropertyGroup>
 	<ItemGroup>
 		<PackageReference Include="System.IO.Ports" Version="10.0.9" />
-		<PackageReference Include="Daqifi.Core" Version="0.23.0" />
+		<PackageReference Include="Daqifi.Core" Version="0.24.0" />
 		<PackageReference Include="Roslynator.Analyzers" Version="4.15.0">
 			<PrivateAssets>all</PrivateAssets>
 			<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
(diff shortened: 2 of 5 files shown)


tylerkron:
**Re: Qodo review findings** — declining both; rationale below. Both concern the intentional no-op design of `BootloaderSessionStreamingDeviceAdapter`, and both were previously evaluated and resolved the same way in #400.

**1. Silent no-op device controls (🐞 Bug)** — keeping as no-ops.
- This adapter is a minimal, inert stand-in for the manual bootloader-update dialog. Its class summary documents that *"serial operations are intentionally no-op,"* and `StartStreaming`/`StopStreaming`/`Send` are already empty no-ops. The 8 new `IStreamingDevice` members (added by Core 0.24.0 #250) extend that same established pattern.
- There is no reachable call site: the adapter is constructed only in `FirmwareDialogViewModel` and passed solely to `IFirmwareUpdateService.UpdateFirmwareAsync`, which drives the bootloader/HID flashing transport and never invokes channel/DIO/DAC/reboot operations. The "silent failure" path isn't exercised in the firmware-update flow.
- Throwing from only the new members would make the class internally inconsistent (some unsupported ops throw, others no-op). Converting the whole adapter to throw/log is a deliberate design change for its own PR rather than a dependency bump — and was already declined for this adapter in #400.

**2. No XML docs for no-ops (📘 Rule 244813)** — keeping consistent with the class.
- No member of this adapter currently carries XML docs; the 8 new members share a single explanatory block comment describing why they are no-ops. Documenting only the new members would be inconsistent with the rest of the type, and an XML-docs request for this same adapter was previously declined in #400.

Net: this PR intentionally extends the adapter's existing, documented no-op pattern to satisfy the new Core 0.24.0 interface surface, with no behavior change to the firmware-update flow.

PR #634, opened 2026-06-23
feat(firmware): unify firmware UX + bootloader download-latest + flash reliability

cptkoolbeenz:
## Summary

Re-homes this PR onto current `main` (the post-#592/#601 architecture) and reframes the feature around a single user-facing idea: **"the firmware."** Users shouldn't need to know a Nyquist device carries both a PIC32 and a separately-flashable WINC1500 WiFi module — to them it's just firmware.

### Why this was reworked

The original branch was cut **before the #592 refactor wave**: before [#601](https://github.com/daqifi/daqifi-desktop/pull/601) extracted the firmware flow out of `DaqifiViewModel` into `FirmwareUpdateCoordinator` (behind `IFirmwareUpdateHost`), and before [#587](https://github.com/daqifi/daqifi-desktop/pull/587) delegated the WiFi version check to Core's `CheckWifiFirmwareStatusAsync`. As a result the original diff added ~900 lines back into `DaqifiViewModel`, re-implemented logic main already ships, and was `CONFLICTING`.

Critically, **main already updates the WiFi module as part of a single "Update Firmware" action** (the coordinator flashes PIC32 → then WiFi, using Core for the version decision). So the original PR's separate WiFi surface was largely redundant with what shipped to main while this branch was open. This rework keeps the genuinely-additive parts and drops the redundant ones.

### What's included

**Bootloader recovery dialog — "just flash the latest firmware":**
- A **DAQiFi Firmware dropdown** that downloads + flashes the latest published firmware, so recovery no longer requires hunting for a `.hex` (browse remains the fallback). Dark combobox styling; Upload is gated until there's something to flash.
- `IFirmwareUpdateService` / `IFirmwareDownloadService` resolved from DI (fail-fast) instead of newing up an `HttpClient`-backed service.

**Flash reliability (internal, no UX change):**
- **Never strand the device** on a failed WiFi flash: when the managed connection is already gone, `ResetLanAfterUpdate`'s transparent-mode exit silently no-ops, so we also send `SYSTem:USB:SetTransparentMode 0` over a raw serial write. Guarded on `!IsConnected` (no happy-path change) and run off the calling thread.
- **Pause + drain all device discovery** (WiFi + serial + HID) for the duration of a bootloader flash, restarting it when the dialog closes — a discovery probe re-opening the HID/serial handle mid-flash produced intermittent `Connecting`-state failures.

### Intentionally dropped (redundant with main)

The separate WiFi **badge / version line / notification / "Update WiFi" button** and the desktop-side hardcoded version check. Main's single "Update Firmware" already covers the WiFi module via the coordinator + Core, so these are unnecessary and contrary to the "just the firmware" model. The debug-only WiFi force-flash tooling for QA can come back as a small follow-up if still needed (it now lives most naturally on the coordinator).

## Testing

- New `FirmwareDialogViewModelTests` cover the dropdown load, upload gating, and download-latest vs browsed-`.hex` precedence.
- Fast unit gate green: **458/458** (`TestCategory!=Ui`).
- Not yet re-validated on the bench against hardware — the reliability changes (raw transparent-mode exit, discovery pause) should get a bench pass before merge.

> Note: this supersedes the original commits on the branch (which were built on a stale pre-#592 base). Original authorship preserved via co-author trailer.

### DIFF
--- a/Daqifi.Desktop.Test/ViewModels/FirmwareDialogViewModelTests.cs
+++ b/Daqifi.Desktop.Test/ViewModels/FirmwareDialogViewModelTests.cs
@@ -1,0 +1,165 @@
+using System.Diagnostics;
+using Daqifi.Core.Device;
+using Daqifi.Core.Firmware;
+using Daqifi.Desktop.ViewModels;
+using Moq;
+
+namespace Daqifi.Desktop.Test.ViewModels;
+
+/// <summary>
+/// Covers the bootloader firmware dialog's "download the latest DAQiFi firmware" path: the dropdown
+/// is populated from the latest release, the Upload command is gated until there's something to flash,
+/// a no-file upload downloads the latest .hex before flashing, and a browsed .hex takes precedence
+/// (no download).
+/// </summary>
+[TestClass]
+public class FirmwareDialogViewModelTests
+{
+    private string _tempHexPath = string.Empty;
+
+    [TestInitialize]
+    public void Setup()
+    {
+        // A real file so the dialog's File.Exists guard passes; the flash service is mocked.
+        _tempHexPath = Path.Combine(Path.GetTempPath(), $"daqifi-fw-{Guid.NewGuid():N}.hex");
+        File.WriteAllText(_tempHexPath, ":00000001FF");
+    }
+
+    [TestCleanup]
+    public void Cleanup()
+    {
+        if (File.Exists(_tempHexPath)) { File.Delete(_tempHexPath); }
+    }
+
+    private static Mock<IFirmwareDownloadService> DownloadServiceReturning(FirmwareReleaseInfo? latest)
+    {
+        var download = new Mock<IFirmwareDownloadService>();
+        download
+            .Setup(s => s.GetLatestReleaseAsync(It.IsAny<bool>(), It.IsAny<CancellationToken>()))
+            .ReturnsAsync(latest);
+        return download;
+    }
+
+    private static FirmwareReleaseInfo Release(int major, int minor, int patch, string tag) => new()
+    {
+        Version = new FirmwareVersion(major, minor, patch, null, 0),
+        TagName = tag,
+        IsPreRelease = false
+    };
+
+    /// <summary>Waits for a fire-and-forget continuation to settle (the constructor's option load).</summary>
+    private static void WaitUntil(Func<bool> condition, int timeoutMs = 2000)
+    {
+        var sw = Stopwatch.StartNew();
+        while (!condition() && sw.ElapsedMilliseconds < timeoutMs)
+        {
+            Thread.Sleep(10);
+        }
+    }
+
+    [TestMethod]
+    public void Constructor_LoadsLatestFirmwareIntoDropdown()
+    {
+        var update = new Mock<IFirmwareUpdateService>();
+        var download = DownloadServiceReturning(Release(3, 6, 1, "v3.6.1"));
+
+        var vm = new FirmwareDialogViewModel("TestHid", update.Object, download.Object);
+
+        WaitUntil(() => vm.AvailableFirmwares.Count > 0);
+        Assert.AreEqual(1, vm.AvailableFirmwares.Count);
+        Assert.IsNotNull(vm.SelectedFirmware);
+        Assert.AreEqual("3.6.1", vm.SelectedFirmware!.Version);
+        Assert.AreEqual("DAQiFi — 3.6.1", vm.SelectedFirmware.Display);
+    }
+
+    [TestMethod]
+    public void Constructor_NoRelease_LeavesDropdownEmpty()
+    {
+        var update = new Mock<IFirmwareUpdateService>();
+        var download = DownloadServiceReturning(null);
+
+        var vm = new FirmwareDialogViewModel("TestHid", update.Object, download.Object);
+
+        // Nothing to select; the upload command must be disabled until the user browses a file.
+        Assert.AreEqual(0, vm.AvailableFirmwares.Count);
+        Assert.IsNull(vm.SelectedFirmware);
+        Assert.IsFalse(vm.UploadFirmwareCommand.CanExecute(null));
+    }
+
+    [TestMethod]
+    public void CanUploadFirmware_TrueOnceAFileIsBrowsed()
+    {
+        var update = new Mock<IFirmwareUpdateService>();
+        var download = DownloadServiceReturning(null);
+        var vm = new FirmwareDialogViewModel("TestHid", update.Object, download.Object);
+
+        Assert.IsFalse(vm.UploadFirmwareCommand.CanExecute(null), "Disabled with no file and no dropdown selection.");
+
+        vm.FirmwareFilePath = _tempHexPath;
+
+        Assert.IsTrue(vm.UploadFirmwareCommand.CanExecute(null), "Enabled once a .hex is browsed.");
+    }
+
+    [TestMethod]
+    public async Task Upload_NoFileBrowsed_DownloadsLatestThenFlashes()
+    {
+        var update = new Mock<IFirmwareUpdateService>();
+        update
+            .Setup(s => s.UpdateFirmwareAsync(
+                It.IsAny<IStreamingDevice>(),
+                It.IsAny<string>(),
+                It.IsAny<IProgress<FirmwareUpdateProgress>>(),
+                It.IsAny<CancellationToken>()))
+            .Returns(Task.CompletedTask);
+
+        var download = DownloadServiceReturning(Release(3, 6, 1, "v3.6.1"));
+        download
+            .Setup(s => s.DownloadLatestFirmwareAsync(
+                It.IsAny<string>(), true, It.IsAny<IProgress<int>?>(), It.IsAny<CancellationToken>()))
+            .ReturnsAsync(_tempHexPath);
+
+        var vm = new FirmwareDialogViewModel("TestHid", update.Object, download.Object);
+        // Stand in for the async dropdown load so we exercise the no-file path deterministically.
+        vm.SelectedFirmware = new Models.FirmwareOption { DeviceModel = "DAQiFi", Version = "3.6.1" };
+
+        await vm.UploadFirmwareCommand.ExecuteAsync(null);
+
+        download.Verify(s => s.DownloadLatestFirmwareAsync(
+            It.IsAny<string>(), true, It.IsAny<IProgress<int>?>(), It.IsAny<CancellationToken>()), Times.Once);
+        update.Verify(s => s.UpdateFirmwareAsync(
+            It.IsAny<IStreamingDevice>(), _tempHexPath, It.IsAny<IProgress<FirmwareUpdateProgress>>(),
+            It.IsAny<CancellationToken>()), Times.Once);
+        Assert.IsTrue(vm.IsUploadComplete);
+        Assert.IsFalse(vm.HasErrorOccured);
+    }
+
+    [TestMethod]
+    public async Task Upload_FileBrowsed_TakesPrecedenceAndDoesNotDownload()
+    {
+        var update = new Mock<IFirmwareUpdateService>();
+        update
+            .Setup(s => s.UpdateFirmwareAsync(
+                It.IsAny<IStreamingDevice>(),
+                It.IsAny<string>(),
+                It.IsAny<IProgress<FirmwareUpdateProgress>>(),
+                It.IsAny<CancellationToken>()))
+            .Returns(Task.CompletedTask);
+
+        var download = DownloadServiceReturning(Release(3, 6, 1, "v3.6.1"));
+
+        var vm = new FirmwareDialogViewModel("TestHid", update.Object, download.Object)
+        {
+            FirmwareFilePath = _tempHexPath
+        };
+
+        await vm.UploadFirmwareCommand.ExecuteAsync(null);
+
+        download.Verify(s => s.DownloadLatestFirmwareAsync(
+            It.IsAny<string>(), It.IsAny<bool>(), It.IsAny<IProgress<int>?>(), It.IsAny<CancellationToken>()),
+            Times.Never);
+        update.Verify(s => s.UpdateFirmwareAsync(
+            It.IsAny<IStreamingDevice>(), _tempHexPath, It.IsAny<IProgress<FirmwareUpdateProgress>>(),
+            It.IsAny<CancellationToken>()), Times.Once);
+        Assert.IsTrue(vm.IsUploadComplete);
+    }
+}
--- a/Daqifi.Desktop/Device/Firmware/FirmwareUpdateCoordinator.cs
+++ b/Daqifi.Desktop/Device/Firmware/FirmwareUpdateCoordinator.cs
@@ -382,8 +382,63 @@ public class FirmwareUpdateCoordinator
         {
             if (lanUpdateModeEnabled)
             {
-                serialStreamingDevice.ResetLanAfterUpdate();
+                // Bring the device back out of WiFi-update / USB-transparent (bridge) mode. The managed
+                // ResetLanAfterUpdate routes through the Core device; if a failed flash has already torn
+                // that connection down, its SetTransparentMode 0 silently no-ops and the device is left
+                // stranded in transparent mode — it stops answering SCPI and vanishes from the app until
+                // it is power-cycled.
+                try
+                {
+                    serialStreamingDevice.ResetLanAfterUpdate();
+                }
+                catch (Exception ex)
+                {
+                    _appLogger.Warning($"ResetLanAfterUpdate failed for {serialStreamingDevice.PortName}: {ex.Message}");
+                }
+
+                // Safety net: only when the managed reset could NOT have worked (the Core connection is
+                // gone), also send the transparent-mode exit over a raw serial write so a failed flash
+                // can never strand the device. Run it off the calling thread — it does blocking serial
+                // I/O with short sleeps.
+                if (!coreDevice.IsConnected)
+                {
+                    await Task.Run(() => ExitWifiTransparentModeRaw(serialStreamingDevice.PortName));
+                }
             }
+        }
+    }
+
+    /// <summary>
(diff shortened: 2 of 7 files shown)


cptkoolbeenz:
Pass 1 applied 5 review findings + 3 suggestions. Two suggestions skipped:

- **Add x:Key to the DarkDialog ComboBox styles** (importance 5) — SKIPPED. `DarkDialog.xaml` is a dialog-scoped *implicit-style theme* (its ComboBox, ComboBoxItem, and GroupBox styles are all implicit `TargetType` styles), merged by 3 dialogs that rely on it. Keying only the ComboBox would break that convention and the dialogs depending on it. 
- **TemplateBinding inside the ComboBox template** (importance 6) — SKIPPED. The flagged border is inside the inner `ToggleButton`'s template (`TargetType="ToggleButton"`), so `{TemplateBinding Background}` binds to the ToggleButton's unset Background, not the ComboBox's — it would drop the dark Surface fill and regress the validated dropdown. (The Popup border would bind correctly, but a partial application is more confusing than leaving the working DynamicResource form.)

cptkoolbeenz:
Pass 3: applied 6 suggestions. Skipped 1:

- **Handle cancellation without error** (importance 7) in `FirmwareDialogViewModel.UploadFirmware` — already implemented: there's a `catch (OperationCanceledException)` at line 193 ahead of the generic `catch`. The suggestion's diff omitted that existing catch.

cptkoolbeenz:
Pass 5: applied 3 (incl. reversing the earlier x:Key decline — verified only FirmwareDialog has a ComboBox among the 3 DarkDialog consumers, so opt-in keying is safe). Skipped 1:

- **Replace `_wifiCheckInProgress` with `SemaphoreSlim`** (importance 4) — declined. The flag is set/read on the UI thread (the probe is fire-and-forget from the ConnectedDevices change handler), so a bool guard is sufficient. Qodo's own rationale notes its proposed diff is incomplete (it doesn't actually guard `CancelWifiFirmwareCheck`).

cptkoolbeenz:
Pass 8: applied the SelectedFirmware XML doc. Skipped 1:

- **FallbackValue/TargetNullValue on WiFi Visibility bindings** (importance 6) — declined. DevicesPanePrototype uses bare `Visibility` bindings throughout (dozens), and `BoolToVis` already returns Collapsed for null, so adding these attributes to only the 4 new WiFi bindings introduces an inconsistent pattern with no behavioral change.

cptkoolbeenz:
Pass 10: applied the constructor XML docs. Skipped 1:

- **StaticResource → DynamicResource for DarkComboBox** (importance 5) — declined. DarkDialog.xaml is merged into FirmwareDialog's own resource dictionary (loaded before content), so the keyed style resolves at parse time with no load-order risk. StaticResource matches the codebase convention for style references and is more performant.

cptkoolbeenz:
Pass 11: applied the fire-and-forget probe try/catch and the dropdown scrollbar visibility. Skipped 1:

- **Implement IDisposable on FirmwareDialogViewModel** (importance 6) — declined. The dialog is shown modally via the dialog service, which doesn't dispose the view-model, so a Dispose() would never be invoked here. `_updateCts` is already cancelled/disposed in UploadFirmware's `finally`, and `LoadFirmwareOptionsAsync` is a short one-shot fetch — so this would be ceremonial churn with no teardown actually wired to it.

cptkoolbeenz:
## Qodo cycle converged (13 passes)

**Bug surface: 0 bugs** on the latest commit. Across the cycle Qodo found and we fixed real issues — off-thread `ConnectedDevices` races, an uncancelable probe CTS, a false upload error, a finder/CTS leak, a stale notification, case-sensitivity bugs, a dropdown virtualization regression, and telemetry visibility — plus MVVM/DI convention compliance.

**Post-convergence verification:**
- Unit tests: **355/355** (zero regressions vs. pre-cycle baseline)
- Full Qodo-driven diff re-reviewed for semantic drift on the high-risk paths (WiFi-probe concurrency, discovery start/stop/restart lifecycle, upload-command gating) — clean.

**Pass 13 remaining suggestions declined (documented dismissal):**
- *Make UI marshaling non-blocking* (imp 6) — `Invoke` is appropriate for a trivial UI-bound update; Qodo's proposed `InvokeAsync`/`ApplyResultAsync` form generates compiler warnings (per its own note). 
- *Guard dispatcher snapshot failures* (imp 4) — already covered by the fire-and-forget `try/catch` wrapping `CheckWifiFirmwareCoreAsync` (the snapshot included), which logs at Error. 

Stopping the cycle here: bug surface clean, tests green, remaining items are regenerating best-practice churn.

cptkoolbeenz:
@tylerkron — requesting your review when you have a chance. 🙏

**What this adds:** end-to-end WiFi (WINC1500) firmware updating + a bootloader-mode firmware dropdown, plus flash-reliability fixes uncovered while batch-programming devices:
- Power on the WINC before the version check after a PIC32 reflash (was needlessly re-flashing current modules), `force` flag so manual WiFi flashes actually run, and a settle so the first attempt is reliable
- Pause + drain all discovery loops during a bootloader flash (HID/serial probes were colliding with the flash)
- Firmware dropdown + dark combobox, accent-styled FLASH WIFI button, and device-vs-WiFi-firmware labels in the success/error dialogs

**Status:**
- Unit tests: **355/355** (zero regressions)
- Ran a full Qodo convergence cycle (13 passes → 0 bugs); it caught several real concurrency/MVVM issues that are all fixed. Skips are documented inline on the thread.
- Self-review + comment hygiene clean.

**One caveat for the merge:** the on-hardware feature validation (bootloader-dropdown flash, auto-WiFi-after-PIC32, manual WiFi force/first-click) is still being run on a device — code review can proceed in parallel, but I'd hold the actual merge until that's confirmed.

Happy to walk through any of the flash paths or the Qodo decisions if useful.

tylerkron:
C:/Program Files/Git/agentic_review

tylerkron:
## Bench test on a real device (Nq1, USB) — mixed results, honest writeup

Ran this branch's Debug build against a physically attached **Nyquist 1 (Nq1, COM3, S/N 9090…4104)**. Some of this validated cleanly; one path failed and is called out below. Not stamping a blanket ✅.

### What passed on hardware

- **Connect + discovery refactor (change C).** Discovered the Nq1 on COM3 and connected cleanly; reconnected again after recovery. The `ConnectHid` HID-discovery drain ran and released the handle with **no drain-timeout/leak warnings** in the log.
- **Main firmware update (coordinator path).** The normal drawer "Update Firmware" flashed **PIC32 3.6.0 → 3.6.2** end-to-end and showed the success dialog. Log confirms my edited `UpdateWifiModuleAsync` ran correctly: `WiFi chip info query succeeded … 19.7.7` → `already up to date (device 19.7.7, latest 19.7.7); skipping WiFi flash`. No errors.
- **Download-latest dropdown UI (change A).** Forcing the device into bootloader (`SYSTem:FORceBoot`), the dialog opened with the **"DAQiFi — 3.6.2"** dropdown populated, dark combobox styling, and the browse fallback. The no-file Upload path *downloaded* the latest `.hex` into the new `…\AppData\Local\DAQiFi\Firmware\PIC32\` dir (the new `GetFirmwareDownloadDirectory()` + `DownloadLatestFirmwareAsync`).

### What did NOT pass

- **The recovery-dialog HID flash did not complete.** On the force-bootloadered device the flash failed at `Connecting to HID bootloader` / `PreparingDevice`, *before* any erase/write. The device was left in bootloader and **recovered via a USB power-cycle** (the 3.6.2 firmware was intact; it booted straight back and now reports 3.6.2). So: the dropdown UI + download worked, but I did **not** get a clean end-to-end recovery flash through the new dialog on this bench.

### My analysis of the failure (hypothesis, not confirmed)

I believe this is **pre-existing and not caused by this PR**, but I have **not proven it** — I did not re-run the same scenario on `main`:
- The flash *call* in `FirmwareDialogViewModel.UploadFirmware` is unchanged from `main` (`_firmwareUpdateService.UpdateFirmwareAsync(_coreDevice, …)`), and the DI-resolved `IFirmwareUpdateService` uses the same `CreateBootloaderHidTransport()` + widened timeouts as `main`'s old fallback (App.xaml.cs:97-104).
- Change C's discovery drain worked (no leak/timeout in logs), and it failed identically with a **single** app instance — so it's not contention and not the drain.
- The drawer flash *succeeded* via the same Core HID flasher (Core orchestrates reboot→connect→flash); only the "connect to an already-in-bootloader device" path failed.

**Next step to confirm:** reproduce on `main` (force bootloader → recovery flash). If it also fails there, it's a Core HID-connect issue on the recovery path (worth a separate issue, possibly relevant to the `0.24.0` bump in #633) rather than anything this PR changed. I'll treat that as an open follow-up, not a claim.

### Net

The behavior this PR actually adds — the **download-latest dropdown** (UI + download) and the **discovery-drain reliability** — exercised correctly on hardware. The end-to-end recovery flash is an **open item** I couldn't close on the bench and haven't yet attributed with certainty. Fast unit gate remains green (458/458).

🤖 Bench validation via [Claude Code](https://claude.com/claude-code)

tylerkron:
## Bench validation (real Nq1, 2026-06-24)

Tested firmware upload on attached hardware (Nq1, COM3). Findings, including what works and an open item:

**What works**
- The normal **"Update Firmware"** drawer flow (the unified UX this PR is about) flashed PIC32 firmware end-to-end on the bench earlier in the session (3.6.0 → 3.6.2). The download-latest + unified-UX parts of the PR are unaffected by the issue below.

**What I found — and root-caused**
- The separate **bootloader recovery dialog** (Connect → Firmware tab → Open Firmware → Upload, against a device already in HID bootloader mode) failed to flash. I reproduced it on the bench and traced it to a **bootloader firmware bug**, not a host bug:
  - The PIC32 HID bootloader never re-arms its USB endpoint after a host-triggered reset/deconfigure — its datastream state machine leaves `readRequest` stale, so the endpoint goes **dark** and the next flash fails on its first write (`HID write failed`) or first read (`Timed out waiting for HID data`).
  - The **trigger** is ordinary host HID handle open/close churn: enumerating the device to read its USB string descriptors opens the handle. The desktop's old discovery did this on the Firmware tab.
  - The dark state **persists until a power-cycle** and is **not recoverable host-side** — confirmed by isolating the flash through `daqifi-core` alone: a fresh bootloader with minimal opens completes, but once darkened even Core's own clean connect can't recover it.
  - Filed as the root-cause fix: **daqifi/daqifi-nyquist-firmware#568**.

**What this PR now does about it (commit 571ab7f)**
- Adds `BootloaderHidFinder` — a zero-open HID discovery for the Firmware tab that enumerates by VID/PID + DevicePath and **never opens the handle**, replacing Core's `HidDeviceFinder` for desktop discovery. This removes the desktop's contribution to the trigger (the extra open that tipped the bootloader dark on the bench) and brings the flow down to Core's minimal flash-flow opens, which complete on a fresh bootloader. The prior "stop scanning after first find" workaround is reverted since continuous polling is now safe.
- Covered by 6 new unit tests; full desktop suite green (464/464).

**Honest scope / open item**
- This is a **host-side mitigation**, not a guaranteed fix. While the firmware bug exists, the flasher itself must open the handle to program, so some residual risk remains — the **definitive fix is firmware (#568)**.
- **End-to-end bench re-validation of the recovery flash is still pending**: the test device is currently sitting in the dark bootloader state from the repro and needs a **power-cycle** to recover (it is *not* bricked — the flash never reached erase, so the app firmware is intact and it boots normally). I'll confirm the recovery flash runs to completion with the mitigation once the device is power-cycled.

tylerkron:
C:/Program Files/Git/agentic_review

tylerkron:
## Bench re-validation update — mitigation did NOT fix the recovery flash

Correcting my earlier comment, which left the impression the host mitigation (commit `571ab7f`) would likely work pending validation. It was bench-tested and **the recovery flash still fails** — identically to before:

```
state Connecting -> Failed
TimeoutException: Timed out waiting for HID data after 30000 ms
  at FirmwareUpdateService.RequestBootloaderVersionAsync -> HidLibraryTransport.ReadAsync
```

i.e. write succeeds, the bootloader version-response read times out at ~10% (Connecting) — the same darkened-endpoint signature. So the zero-open discovery, on its own, **does not resolve the issue**.

What this means:
- **Root cause remains the bootloader firmware** (daqifi/daqifi-nyquist-firmware#568). That is the only reliable fix.
- The mitigation in `571ab7f` is at best *defense-in-depth* (it removes the desktop's extra pre-flash handle open) — **not** a fix. It should be treated as such, not relied on, and possibly reverted/reframed.
- One isolation experiment is still open to decide its fate: re-run the Core CLI against a **fresh** bootloader with no pre-enum vs. with the desktop's exact `GetHidDevices`+`DevicePath` enumeration. That tells us whether (a) a clean host flow reliably flashes a fresh bootloader and the desktop has a further difference, or (b) the firmware bug trips on Core's own unavoidable opens and no host change can guarantee success. This needs bench time on a recovered device and is not yet done.

Until that experiment runs, **do not treat the recovery-dialog firmware flash as fixed by this PR.** The unified-UX and download-latest parts of the PR are unaffected and were validated separately.

tylerkron:
## Isolation experiment complete — mitigation reverted

Closing the loop from the previous comment. I ran the decisive isolation test: the **clean Core CLI flow** (no desktop, no pre-flash enumeration) against a freshly `FORceBoot`-entered bootloader. It **also failed** with the same `TimeoutException` on the version-response read.

That settles it:
- The recovery flash fails **probabilistically even via the minimal CLI path**, triggered by **Core's own unavoidable handle opens** during the flash — not by anything the desktop adds.
- A host-side mitigation therefore **cannot** make it reliable. The zero-open discovery I'd pushed (`571ab7f`) had no effect on the bench.

**Action taken:** reverted `571ab7f` — this branch is force-pushed back to the unified-UX + download-latest + HID-timeout-widening work only (`e3a440e`). No `BootloaderHidFinder`, no added HidSharp dependency. Build green.

The recovery-flash reliability fix lives in firmware: **daqifi/daqifi-nyquist-firmware#568** (re-arm the bootloader endpoint on USB reset/deconfigure), now updated with this evidence. The parts of this PR that *were* validated on the bench (the normal "Update Firmware" drawer flow flashing PIC32 3.6.0 → 3.6.2, and the unified UX) stand unaffected.

cptkoolbeenz:
## Hardware testing surfaced a manufacturing-workflow conflict + a few flash bugs

We ran this branch against real devices today (programming/QA). The "unify everything as *the firmware*" direction reads well, but it conflicts with how we actually program units on the line, and a couple of the flash paths regressed. Writing it up here so we can decide the direction together rather than me quietly re-adding things.

### The manufacturing workflow this needs to support

Units are programmed in two stages:
1. **PIC32 bootloader + firmware are flashed at the bench** (jig/ICSP), and it's generally already the current firmware.
2. The unit is then connected to the **desktop app only to update the WiFi (WINC) firmware.**

Two hard requirements fall out of this:
- **We must be able to flash the WiFi module on its own, without reflashing the PIC32.** The unified "Update Firmware" reflashes PIC32 too, which is wasted time **per unit** on the line.
- **We must see the WiFi FW version in the UI.** Operators read it to (a) decide whether a flash is needed and (b) verify a flash actually took — WiFi flashes sometimes fail and we retry. While we're debugging *and* shipping at the same time, that operator-visible readback is essential.

So the **WiFi version readout** and a **WiFi-only flash** aren't redundant with the unified update or a "debug-only" nicety — for production they're the **primary** path. The PR description punts the debug WiFi force-flash to "a follow-up if still needed (it now lives most naturally on the coordinator)"; this is that signal — it's needed, and it should be first-class, not debug-gated.

### Flash bugs we hit on this branch (with evidence)

1. **Failed WiFi flash strands the device in USB-transparent / FW-update mode.** The cleanup sends `SYSTem:USB:SetTransparentMode 0` over the managed connection, but in transparent mode the device ignores protobuf *and* that connection still holds the port — so the exit no-ops and the device vanishes from the app (and won't re-connect, won't enter the bootloader, sits in standby). Recovery required a **raw** `SetTransparentMode 0` after releasing the port. A failed flash must never strand a unit.

2. **False success.** When the device doesn't hand the port to the WINC tool, the tool exits with **no output** and the flash reports **success**. Core only fails on `"Programming device failed"` + nonzero exit (`FirmwareUpdateService.cs:439-447`), so a no-output exit-0 run slips through. The fix is a **positive** success token — see below.

3. **No actionable failure guidance** — operators get a generic message instead of "disconnect / power-cycle / retry."

### Proper success oracle (from the tool's real output)

The WINC tool does write-with-verify. Across today's logs the verdict is unambiguous:
- **`verify passed`** appears exactly on the **successful** flashes (`begin write operation → begin verify operation → verify passed`).
- Every failure shows `Programming device failed` and **no** `verify passed`.

So success should require the **positive** token, not just the absence of a failure string:
```
SUCCESS iff output contains "verify passed"
FAILURE if "Programming device failed" | "error: invalid chip ID" |
           "number of sectors failed verification" | "NACK received" |
           "unable to perform write operation when not connected" | "error: failed to …"
        OR exit != 0
        OR "verify passed" never appeared   (covers the no-output false success)
```
(Separately, `chip ID 0x00050000` vs the good `0x001503a0` is a bad/un-bootloadered WINC — hardware, not parsing — but worth surfacing to the operator too.)

### Proposal

Keep the unified "Update Firmware" for end users, but **restore as first-class, on `FirmwareUpdateCoordinator`:**
- **WiFi FW version readout** in the UI.
- **WiFi-only flash** (no PIC32 reflash).
- **`verify passed`-based success parsing.**
- **Guaranteed transparent-mode exit** on failure (raw, after releasing the port).
- **Brief failure guidance.**

We have all of this working on a session backup branch (`backup/wifi-fixes-session-2026-06-24`) on the old architecture and can port it onto the coordinator. Happy to do that work or hand it over — mainly want to align on keeping the WiFi version + WiFi-only flash, since that's a real production constraint.

Related: filed #638 for the stale-connection-status gap (desktop never subscribes to Core's `StatusChanged`), which is what made the stranded/standby units show as "connected" and reuse a dead connection. That one is independent of this PR's architecture.

PR #639, opened 2026-06-25
feat(firmware): WiFi module version readout, standalone WiFi flash, flash reliability

cptkoolbeenz:
## Why

The merged WiFi firmware flow (#634) re-homed the firmware update logic onto `FirmwareUpdateCoordinator`, but in the process it dropped three WiFi features that the original implementation had and that the manufacturing line depends on. This PR re-homes them onto the coordinator architecture.

The PIC32 bootloader flash itself was **not** regressed by the re-home — its desktop/Core code path is equivalent to the working version (the intermittent flash failures trace to the firmware-side dark-endpoint issue, daqifi-nyquist-firmware#568, which needs a cold power-cycle). This PR is purely the missing WiFi features + WiFi-flash reliability.

## What

**1. WiFi module firmware version readout**
- `IStreamingDevice.WifiFirmwareVersion`, `HasWincWifiModule` (Nyquist/WINC1500 only — ESP32/unknown devices integrate WiFi into the SoC and are skipped), `IsWifiFirmwareOutdated`.
- A connect-time probe powers on the WINC (`SYSTem:POWer:STATe 1`), reads chip info with startup retries, and flags modules below the `19.7.7` floor — or unreadable — as needing a flash. Surfaced on the device pane (`WIFI MODULE VERSION`) and tiles.

**2. Standalone WiFi-only flash**
- `UpdateWifiFirmwareOnly` command flashes just the WINC, independent of the PIC32. The line flashes bootloader+firmware at the bench, then updates only WiFi from the app — saving a full firmware reflash per unit.

**3. WiFi-flash reliability**
- WiFi-update-mode **settle delay** before the flash tool runs (first attempt otherwise exits without programming).
- **Abort** when the device won't enter LAN update mode, instead of running the tool against an unready device.
- **False-success guard**: a flash that returns faster than a real WINC program (≈tens of seconds) never actually programmed — surfaced as a failure with power-cycle guidance rather than a bogus success.
- **Robust transparent-mode exit**: a raw `SYSTem:USB:SetTransparentMode 0` that releases the held managed port and retries, so a failed flash never strands the device in transparent mode.
- The coordinator **cancels the in-flight connect-time WiFi probe before any PIC32 flash** so the probe's SCPI exchange can't corrupt the bootloader handshake.

## Architecture

- Device-layer state (`WifiFirmwareVersion`/`HasWincWifiModule`/`IsWifiFirmwareOutdated`, `PowerOnWifiModule`) lives on `IStreamingDevice`/`AbstractStreamingDevice`/`SerialStreamingDevice`.
- Flash mechanics (version check, download, LAN-mode prep, false-success guard, transparent exit) live in `FirmwareUpdateCoordinator`; the WiFi-flash core (`UpdateWifiModuleAsync`/`FlashWifiPackageAsync`) is shared by the auto post-PIC32 path and the standalone command.
- The connect-time probe, notifications, and the `UpdateWifiFirmwareOnly` command live in `DaqifiViewModel` (where the dispatcher + notification list + connected-device collection are); the coordinator reaches back through `IFirmwareUpdateHost.CancelWifiFirmwareProbe()`.

## Testing

- New: `WifiFirmwareNeedsFlash` version-decision coverage (null/empty/below/at/above/unparseable) + a false-success-guard test (an implausibly fast flash fails and still exits transparent mode).
- Updated existing firmware-flow tests to the chip-info contract.
- Fast gate green: **465 tests** (`TestCategory!=Ui`). The settle delay + false-success guard are injectable so unit tests stay fast; production defaults (5 s / 15 s) are unchanged and exercised by the integration gate.

### DIFF
--- a/Daqifi.Desktop.DataModel/Daqifi.Desktop.DataModel.csproj
+++ b/Daqifi.Desktop.DataModel/Daqifi.Desktop.DataModel.csproj
@@ -8,7 +8,7 @@
 	</PropertyGroup>
 	<ItemGroup>
 	  <PackageReference Include="CommunityToolkit.Mvvm" Version="8.4.2" />
-	  <PackageReference Include="Daqifi.Core" Version="0.24.0" />
+	  <PackageReference Include="Daqifi.Core" Version="0.24.1" />
 	  <PackageReference Include="Roslynator.Analyzers" Version="4.15.0">
 		<PrivateAssets>all</PrivateAssets>
 		<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
--- a/Daqifi.Desktop.DataModel/Daqifi.Desktop.DataModel.csproj
+++ b/Daqifi.Desktop.DataModel/Daqifi.Desktop.DataModel.csproj
@@ -8,7 +8,7 @@
 	</PropertyGroup>
 	<ItemGroup>
 	  <PackageReference Include="CommunityToolkit.Mvvm" Version="8.4.2" />
-	  <PackageReference Include="Daqifi.Core" Version="0.24.1" />
+        <PackageReference Include="Daqifi.Core" Version="0.24.1" />
 	  <PackageReference Include="Roslynator.Analyzers" Version="4.15.0">
 		<PrivateAssets>all</PrivateAssets>
 		<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
--- a/Daqifi.Desktop.IO/Daqifi.Desktop.IO.csproj
+++ b/Daqifi.Desktop.IO/Daqifi.Desktop.IO.csproj
@@ -7,7 +7,7 @@
 	</PropertyGroup>
 	<ItemGroup>
 		<PackageReference Include="System.IO.Ports" Version="10.0.9" />
-		<PackageReference Include="Daqifi.Core" Version="0.24.0" />
+		<PackageReference Include="Daqifi.Core" Version="0.24.1" />
 		<PackageReference Include="Roslynator.Analyzers" Version="4.15.0">
 			<PrivateAssets>all</PrivateAssets>
 			<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
--- a/Daqifi.Desktop.IO/Daqifi.Desktop.IO.csproj
+++ b/Daqifi.Desktop.IO/Daqifi.Desktop.IO.csproj
@@ -7,7 +7,7 @@
 	</PropertyGroup>
 	<ItemGroup>
 		<PackageReference Include="System.IO.Ports" Version="10.0.9" />
-		<PackageReference Include="Daqifi.Core" Version="0.24.1" />
+        <PackageReference Include="Daqifi.Core" Version="0.24.1" />
 		<PackageReference Include="Roslynator.Analyzers" Version="4.15.0">
 			<PrivateAssets>all</PrivateAssets>
 			<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
--- a/Daqifi.Desktop.Test/ViewModels/DaqifiViewModelFirmwareUpdateTests.cs
+++ b/Daqifi.Desktop.Test/ViewModels/DaqifiViewModelFirmwareUpdateTests.cs
@@ -187,7 +187,11 @@ public class DaqifiViewModelFirmwareUpdateTests
         await coordinator.UploadFirmwareAsync();
 
         Assert.AreSame(coreDevice, pic32Device);
-        Assert.AreSame(coreDevice, wifiDevice);
+        // Unlike the PIC32 flash, the WiFi flash releases the managed serial connection and runs the
+        // WINC tool against a no-op device (the tool reaches the board over raw serial via bridge
+        // activation) — that's what frees the COM port so the tool can open it. So the WiFi flash does
+        // NOT reuse the live core device.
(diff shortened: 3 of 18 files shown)


cptkoolbeenz:
Qodo pass 1 responses (896eec6):

1. **XML docs on `IsWifiFirmwareUpdate`** — ✅ fixed.
2. **`Thread.Sleep` in coordinator** — kept (disagree). These are in the synchronous raw-serial callbacks (bridge-activation `Action`, transparent-mode exit) that run off the UI thread via `Task.Run`; they're short, deliberate serial-port timing waits (the device needs ~ms between DTR/commands), not UI-blocking, and the surrounding API is synchronous (`Action`, not async), so `await Task.Delay` isn't available without reshaping the process-runner contract.
3. **Chip-info catch omits exception** — ✅ fixed; now logs the exception object (stack trace) in both the coordinator and the connect-time probe.
4. **Unconditional WiFi flash** — ✅ fixed; `UpdateWifiModuleAsync` now early-returns unless `serialStreamingDevice.HasWincWifiModule`, so ESP32/Unknown devices never get an unsupported WiFi flash.
5. **Probe state unsynchronized** — already resolved.
6. **Non-monotonic flash timing** — ✅ fixed; the false-success guard now uses a monotonic `Stopwatch` instead of `DateTime.UtcNow`.
7. **Constants not SCREAMING_SNAKE_CASE** — kept (disagree). The codebase is mixed, and the adjacent firmware/timing constants (`InitialStatusPollInterval`, `InitialStatusTimeout`, `WifiChipInfoRetryDelay`, etc.) use PascalCase — these match the local convention, so renaming would create local inconsistency.

cptkoolbeenz:
Pass 2 (2f5b66c):
- **Use InvokeAsync instead of Invoke** — ✅ fixed; the probe now `await dispatcher.InvokeAsync(ApplyResult)` so it doesn't block the worker thread.
- **Marshal FirmwareUpdateStatusText** — not changed: it's a simple `[ObservableProperty]` string bound to a TextBlock, and WPF's binding engine already marshals scalar property-change updates to the UI thread (unlike collection mutations, which we do dispatch). Low risk; left as-is.

cptkoolbeenz:
Pass 3 (418bc49):
- **Quiesce before try/finally** — ✅ fixed; moved `QuiesceWifiFirmwareProbeAsync()` inside the try so a fault still runs the finally (no stuck "uploading" state).
- **WiFi command CanExecute stale** — ✅ fixed; re-notify `UpdateWifiFirmwareOnlyCommand` when the firmware drawer opens (HasWincWifiModule derives from the async-populated DeviceType); the probe result also re-notifies.
- **QuiesceWifiFirmwareProbeAsync swallows exceptions** — intentional and commented: the in-flight probe already logs its own faults via `CheckWifiFirmwareAsync`; here we only need it *stopped* before the flash, and its outcome is irrelevant. Narrowing/logging again would just double-log.
- **Thread.Sleep / constants casing** — unchanged (see prior replies: off-UI-thread raw-serial timing; PascalCase matches adjacent firmware constants).

cptkoolbeenz:
This pass (ea030bd):
- **Untracked WiFi probe task** — ✅ fixed; all probe starts route through `TriggerWifiFirmwareProbe`, which tracks `_wifiProbeTask` so the pre-flash quiesce actually awaits the in-flight probe (the brick-safety guarantee).
- **Quiesce ignores upload cancel** — ✅ fixed; `QuiesceWifiFirmwareProbeAsync(CancellationToken)` uses `Task.WaitAsync`, and both coordinator call sites pass the upload token.
- **Probe cancel waits on UI** — ✅ fixed; `dispatcher.InvokeAsync(ApplyResult, …, wifiCheckToken)`.
- **WiFi command CanExecute stale** — ✅ fixed; re-notified in `TriggerWifiFirmwareProbe` (the prior fix lived in the now-deleted `OpenFirmwareUpdateSettings`).
- **Converter XML docs** — ✅ added. **Tabs in csproj** — ✅ normalized the bumped lines to 4-space.
- **Dead code** — removed `OpenFirmwareUpdateSettings` (unbound; the firmware UI moved inline).
- **Thread.Sleep / constants casing / Quiesce swallow** — unchanged (prior rationale: off-UI-thread raw-serial timing; PascalCase matches adjacent firmware constants; the probe already logs its own faults, the quiesce only needs it *stopped*).

Architectural note: per direction to maximize Core reuse, the WiFi-flash orchestration the desktop currently wraps (port-release wait, the no-op-adapter handoff, the false-success guard, LAN-mode prep/recovery, chip-info retry) is captured for delegation to Core in **daqifi-core#269**.

cptkoolbeenz:
Suggestions pass (8a80539):
- **Prevent overlapping probe execution** — ✅ applied; `_wifiCheckInProgress` is now an `Interlocked` single-flight guard.
- **Prevent null binding visibility issues** — ✅ applied; `FallbackValue=Collapsed` on the WiFi visibility MultiBindings (the converter already treats null/unset as false; this makes it explicit).

Declined / deferred with rationale:
- **Avoid probing WiFi on connect** — by design. The connect-time probe is required so the inline WiFi version populates without opening anything, it runs *after* init (never racing the handshake), and it's **debug-gated** so normal users never query the WINC. Removing it reintroduces the "version unknown" bug.
- **Serialize SCPI command sending** — probe-vs-flash SCPI is already serialized by the `IsFirmwareUploading`/`DeviceBeingUpdated` guards + the pre-flash quiesce; `CoreDevice.Send` is the canonical producer path.
- **Marshal UI updates / dispatch property updates** — the cross-thread mutations already go through `dispatcher.InvokeAsync`; scalar `[ObservableProperty]` strings are auto-marshaled by WPF's binding engine.
- **Hardcoded XAML colors** — pre-existing inline-pane styling, out of scope for this PR.

Architectural note: the WiFi-flash orchestration (port handoff, false-success guard, LAN-mode prep/recovery, chip-info retry) is slated to move into Core per **daqifi-core#269**, which will let the desktop drop much of this.

PR #651, opened 2026-06-26
fix(plots): run view-model host-init once so plots render dark

tylerkron:
## Summary

The Live Graph plot area rendered OxyPlot's **white** default while the rest of the app is dark-themed — most visible with no device connected. Root cause is an **inverted one-time host-init guard** in `DaqifiViewModel`. This PR fixes that and hardens every OxyPlot surface against the same failure mode.

## Root cause (proven at runtime)

`DaqifiViewModel`'s init block — which creates the dark-themed `PlotLogger` (`Plotter`) and `DatabaseLogger` (`DbLogger`) and assigns them to the bound properties — was guarded by `if (app.IsWindowInit)` with the flag set *afterwards*:

```csharp
if (app.IsWindowInit)   // runs init only if ALREADY initialized — inverted
{
    ... Plotter = new PlotLogger(); DbLogger = new DatabaseLogger(...); ...
}
app.IsWindowInit = true;
```

The view model is constructed exactly once (`MainWindow.xaml.cs` sets `DataContext = new DaqifiViewModel()`), so on that single construction the flag is `false`, the block is skipped, `Plotter`/`DbLogger` stay null, and `{Binding Plotter.PlotModel}` resolves to null → the `PlotView` falls back to OxyPlot's white default. Verified at runtime: pre-fix the constructor ended with `Plotter == null`; post-fix it is non-null.

The guarded code is byte-identical to `v3.2.0`, so this is a **latent** defect that surfaces whenever the view model is constructed only once. Note the same skipped block also did command registration, the DB logger, the disk-space monitor, profile load, the firewall rule, and the long-lived singleton subscriptions — so the impact was broader than the plot.

## Changes

- **`DaqifiViewModel.cs`** — invert the guard to `if (!app.IsWindowInit)` and set the flag at the top of the block, so one-time init runs on the first/only construction (the instance actually bound to the window) and never retries after a partial failure.
- **`LiveGraphPane.xaml`** — defensive `Background="{StaticResource Surface}"` on the live PlotView.
- **`LoggedDataPanePrototype.xaml`** — same defensive `Background` on the Logged Data main plot and overview minimap. They bind to `DbLogger.PlotModel` / `MinimapPlotModel` (built in the same init block) and shared the identical null-model cause; this keeps every plot surface consistently dark regardless of model/visibility state.

## Scope notes

- These are the only three OxyPlot surfaces in the app (Live Graph ×1, Logged Data ×2). `PlotViewModel` / `PlottingViewModel` are never instantiated (dead code) and are intentionally not touched here.

## Testing

- `dotnet build` — 0 errors.
- Fast unit gate (`TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests`): **463 passed / 0 failed** (Daqifi.Desktop.Test) + **1 passed** (Daqifi.Desktop.IO.Test).
- Root cause confirmed at runtime (`Plotter` null → non-null); empty-state Live Graph renders dark.

### DIFF
--- a/Daqifi.Desktop/View/Prototype/LiveGraphPane.xaml
+++ b/Daqifi.Desktop/View/Prototype/LiveGraphPane.xaml
@@ -357,6 +357,7 @@
             <Grid Grid.Column="0">
                 <oxy:PlotView x:Name="DataLog"
                               Model="{Binding Plotter.PlotModel}"
+                              Background="{StaticResource Surface}"
                               TabIndex="0"
                               Visibility="{Binding IsSdCardLoggingActive, Converter={StaticResource InvertedBoolToVis}}">
                     <oxy:PlotView.DefaultTrackerTemplate>
--- a/Daqifi.Desktop/View/Prototype/LoggedDataPanePrototype.xaml
+++ b/Daqifi.Desktop/View/Prototype/LoggedDataPanePrototype.xaml
@@ -410,6 +410,7 @@
                         <!-- Main plot -->
                         <oxy:PlotView Grid.Row="1"
                                       Model="{Binding DbLogger.PlotModel}"
+                                      Background="{StaticResource Surface}"
                                       Margin="16,8,16,0">
                             <oxy:PlotView.DefaultTrackerTemplate>
                                 <ControlTemplate>
@@ -520,6 +521,7 @@
                         <!-- Overview minimap -->
                         <oxy:PlotView Grid.Row="3"
                                       Model="{Binding DbLogger.MinimapPlotModel}"
+                                      Background="{StaticResource Surface}"
                                       Margin="16,0,16,8"
                                       Height="80">
                             <oxy:PlotView.Style>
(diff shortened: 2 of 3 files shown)


tylerkron:
### Response to Qodo review

**🐞 "Init flag blocks retry" — fixed in `8d260e2`.**

Agreed, good catch. Setting `app.IsWindowInit = true` *before* the `try` meant a throw during init could strand `Plotter`/`DbLogger`/`SummaryLogger` as null while permanently marking init "done" — and `DbLogger` is dereferenced unconditionally via `ILoggingSessionListHost`, so that's a genuine NRE crash path with no in-process recovery.

Fix: moved the flag assignment to the **end of the `try`**, so it's set only after every init step succeeds. A partial failure now leaves the flag clear (a later construction can retry) instead of locking in a half-initialized state; the success path is unchanged (init still runs on the first/only construction). Preferred "set on success" over resetting in `catch` — same intent, single write. Comment updated to match.

**High-level alternatives (from the summary comment):** considered all three — moving init to the App composition root, lazy-initializing the loggers, and a default fallback `PlotModel`. I agree with your own recommendation that none are needed here: the composition-root move is a larger startup refactor with higher regression risk, and lazy-init / fallback-model both risk masking a real init failure (plots appear but don't function). Keeping this PR as the minimal-risk guard fix; the composition-root cleanup is worth a separate issue if we decide to pursue it.

PR #654, opened 2026-06-28
feat(firmware): grab and hold a detected HID bootloader until flash

cptkoolbeenz:
## What

When a bootloader-VID/PID device (`0x04D8/0x003C`) appears in the connection dialog, the app now immediately **grabs and holds** its HID handle until the user flashes it:

- **`BootloaderHoldService`** (DI singleton over the shared exclusive `IHidTransport`): on detection, opens the handle and keeps an **interrupt-IN read continuously pending**. A pending read IRP keeps the USB link active so Windows can't selective-suspend the device — the suspend whose reopen-without-`SET_CONFIGURATION` wedges the bootloader (daqifi-nyquist-firmware#568). Reads carry no payload, so unlike a write they can't be mis-parsed as a stray command.
- **Hand-off**: `PauseForFlashAsync` stops the keep-alive but leaves the handle **open and warm** for the flasher; `ReleaseAsync` closes it on dialog close. Wired into `ConnectionDialogViewModel` (grab on detect, hand off before the flash dialog, re-grab after a cancelled flash, release on close).
- **Discovery** stops repolling once a bootloader is listed, so it isn't re-opened while held.
- **`FirmwareUpdateServiceConfig`** opens the bootloader transport **exclusively** (`ExclusiveAccess`), locking out other openers during the hold + flash (the bootloader's CRC gate is disabled, so a stray frame could be mis-parsed as an ERASE).

## Dependency

Requires **Daqifi.Core 0.26.0** (adds `HidLibraryTransport.ExclusiveAccess` — daqifi-core#274). Bumps `Daqifi.Core` 0.25.0 → 0.26.0.

## Testing

- 5 new `BootloaderHoldService` unit tests + an `ExclusiveAccess` config test. Full fast gate **469/469**.
- **Hardware-validated** on a sitting 1.4 bootloader: keep-alive hold across a 60s idle, then flash all 41,509 records + **4-region CRC verify + JMP** (~24s). The hold must be grabbed *before* the device suspends — it prevents the wedge, it can't revive an already-suspended one (that still needs a power-cycle or the 2.0 firmware re-enum).

### DIFF
--- a/Daqifi.Desktop.DataModel/Daqifi.Desktop.DataModel.csproj
+++ b/Daqifi.Desktop.DataModel/Daqifi.Desktop.DataModel.csproj
@@ -8,7 +8,7 @@
 	</PropertyGroup>
 	<ItemGroup>
 	  <PackageReference Include="CommunityToolkit.Mvvm" Version="8.4.2" />
-        <PackageReference Include="Daqifi.Core" Version="0.25.0" />
+        <PackageReference Include="Daqifi.Core" Version="0.26.0" />
 	  <PackageReference Include="Roslynator.Analyzers" Version="4.15.0">
 		<PrivateAssets>all</PrivateAssets>
 		<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
--- a/Daqifi.Desktop.IO/Daqifi.Desktop.IO.csproj
+++ b/Daqifi.Desktop.IO/Daqifi.Desktop.IO.csproj
@@ -7,7 +7,7 @@
 	</PropertyGroup>
 	<ItemGroup>
 		<PackageReference Include="System.IO.Ports" Version="10.0.9" />
-        <PackageReference Include="Daqifi.Core" Version="0.25.0" />
+        <PackageReference Include="Daqifi.Core" Version="0.26.0" />
 		<PackageReference Include="Roslynator.Analyzers" Version="4.15.0">
 			<PrivateAssets>all</PrivateAssets>
 			<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
--- a/Daqifi.Desktop.Test/Device/Firmware/BootloaderHoldServiceTests.cs
+++ b/Daqifi.Desktop.Test/Device/Firmware/BootloaderHoldServiceTests.cs
@@ -1,0 +1,155 @@
+using System.IO;
+using Daqifi.Core.Communication.Transport;
+using Daqifi.Desktop.Common.Loggers;
+using Daqifi.Desktop.Device.Firmware;
+using Moq;
+
+namespace Daqifi.Desktop.Test.Device.Firmware;
+
+/// <summary>
+/// Verifies the bootloader hold service grabs a sitting HID bootloader, keeps an interrupt-IN read
+/// continuously pending (the keep-alive that prevents the #568 USB selective-suspend wedge), and hands
+/// the handle off to the flasher without disconnecting — versus releasing it outright on close.
+/// </summary>
+[TestClass]
+public class BootloaderHoldServiceTests
+{
+    private Mock<IHidTransport> _transport = null!;
+    private Mock<IAppLogger> _logger = null!;
+    private int _readCount;
+
+    [TestInitialize]
+    public void Setup()
+    {
+        _transport = new Mock<IHidTransport>();
+        _logger = new Mock<IAppLogger>();
+        _readCount = 0;
+
+        _transport
+            .Setup(t => t.ConnectAsync(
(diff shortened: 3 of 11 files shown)


cptkoolbeenz:
**Re: the two rule-violation findings — declining, with rationale.** (Both bugs fixed in 346924f.)

1. **`IsHolding`/members use `<inheritdoc/>` without `<summary>`** — `<inheritdoc/>` *is* valid XML documentation; it inherits the `<summary>` from `IBootloaderHoldService` (which is fully documented), and the build raises no CS1591. Adding explicit summaries would duplicate the interface docs and risk drift. Keeping `<inheritdoc/>` (the standard DRY pattern for interface implementations).

2. **`DefaultKeepAliveReadTimeout` not SCREAMING_SNAKE_CASE** — the project applies SCREAMING_SNAKE_CASE to `const`s (e.g. `BOOTLOADER_VENDOR_ID` here) but PascalCase to `public static readonly` fields, matching the sibling `FirmwareUpdateServiceConfig.BootloaderHidTimeout` (also a `public static readonly TimeSpan`). Renaming to SCREAMING_SNAKE would be inconsistent with that established pattern.

cptkoolbeenz:
**Pass 2 applied** (resume hold on flash failure; `Volatile.Read` for `IsHolding`). **Declining the remaining 3 suggestions:**
- **Document public constructors** — the classes carry type-level summaries and the params are self-describing; the codebase doesn't uniformly XML-doc view-model constructors, so adding them here would be inconsistent.
- **Observe/log background failures** & **Log release task failures** — the fire-and-forget `BeginHoldAsync`/`ReleaseAsync` are best-effort and already catch their own exceptions internally (transport errors are logged and swallowed), so the `_ =` tasks don't fault; there's nothing unobserved to surface.

cptkoolbeenz:
**Pass 3 applied**: bounded the graceful keep-alive stop (so `PauseForFlashAsync` can't hang on a stuck read) and de-flaked the keep-alive-fault test (bounded retry instead of a fixed sleep).

Still declining the re-listed style items — `<inheritdoc/>` is valid/DRY XML doc (build raises no CS1591), and the view-model constructors are covered by type-level summaries with self-describing params (consistent with the rest of the codebase, which doesn't XML-doc VM constructors).

cptkoolbeenz:
**Pass 4**: fixed the cleanup-delay (the failure re-grab is now fire-and-forget, so it can't delay clearing the uploading state). Declining **"Don't block flash on pause"** (importance 3): `PauseForFlashAsync` is already internally best-effort — `StopKeepAliveAsync` catches its own errors — so it won't throw into the flash path.

cptkoolbeenz:
**Pass 5**: bounded the *final* keep-alive wait in `StopKeepAliveAsync` (abandon a non-cancelable read instead of awaiting forever — `PauseForFlashAsync` can no longer deadlock the hand-off), and removed the flash-failure re-grab from `FirmwareDialogViewModel` (it was the unobserved fire-and-forget + delayed cleanup; `ConnectHid`'s `finally` already re-grabs the hold on dialog close, covering the same recovery with less code).

cptkoolbeenz:
**Pass 6 — and calling convergence on the bug side.**

Applied the fix for 'Abandoned keep-alive race' by **reverting** the bounded-abandon logic: that finding correctly showed abandoning a still-running read is *worse* than awaiting it — the orphaned read holds the shared transport's I/O lock, so the flasher's reconnect would race/block on it regardless. `StopKeepAliveAsync` now simply awaits the loop, which is **self-bounded by the transport's per-read timeout (~1s)**: once `_stopKeepAlive` is set, the in-flight read returns/throws and the loop exits within that window (the hard path also cancels). Rationale is in-code.

Pre-empting the earlier 'can still hang' concern: the only way the await hangs is a transport that ignores **both** its read timeout **and** cancellation — a broken transport — and in that case abandoning wouldn't help (the orphaned read still blocks the flasher). So awaiting is the correct choice; there is no better alternative.

**Cycle summary:** genuine bugs fixed — hold-across-the-dialog (was paused too early), stale-hold re-establish, and a flaky test. Reasonable hardening applied (resume-on-failure → later simplified, thread-safe `IsHolding`). The last three passes were Qodo iterating on the *defensive code it asked for*, each refinement spawning the next; this pass simplifies to the correct minimal design. Declining the residual style items (`<inheritdoc/>`, VM-constructor docs) per the rationale above. The hold mechanism itself is unchanged, hardware-validated, and CI-green.

cptkoolbeenz:
Superseded by #656. The app-global `BootloaderWatcher` generalizes and replaces this dialog-scoped, single-device grab-and-hold (which the field test found insufficient) — it holds **every** detected HID bootloader app-wide and flashes any one by device path. #656 carries the full grab-and-hold → multi-device story (all the commits from this branch plus the watcher), so merging it covers everything here. Closing to avoid merging an approach we've moved past.

Issue #655, opened 2026-06-29
Track each device by physical USB location to disambiguate multiple connected devices


cptkoolbeenz:
## Summary

Track each DAQiFi device by its **physical USB location** (hub/port topology path) so the app never loses track of *which physical device is which* when more than one device — especially more than one HID bootloader — is connected at the same time.

## Why now

The new app-global bootloader watcher (multi-device "grab and hold all", see daqifi-nyquist-firmware#568) makes this gap concrete. Multiple PIC32 HID bootloaders are **identical**: same `VID 0x04D8 / PID 0x003C`, and **no serial number** (`GetSerialNumber` throws). The watcher works around this today by addressing each bootloader by its **HID device path**, which is sufficient *while the device stays in bootloader mode*. But a device's identity changes as it moves through its lifecycle:

- **Application/serial mode:** identified by COM port + device serial number.
- **HID bootloader mode:** identified by VID/PID + HID device path, **no serial**.
- **After a flash:** re-enumerates (possibly at a new path/port handle).

There is currently **no stable identity that survives these transitions**, so with multiple devices connected the app cannot reliably answer "the bootloader at this path is the same physical unit that was on COM7 a moment ago."

## Proposed behavior

Maintain a stable per-device identity keyed on the **USB physical location** (the hub/port topology — e.g. the `USB\ROOT_HUB...\\<port chain>` location path, which is stable across mode transitions and uniquely distinguishes physically-distinct but otherwise-identical devices). Use it to:

- Correlate a device across serial ⇄ bootloader ⇄ post-flash re-enumeration.
- Label devices in the UI by physical location when serial/name are unavailable (e.g. "Bootloader on USB port 1-3").
- Disambiguate identical bootloaders without relying solely on the transient HID device path.

## Concrete problem it resolves

`FirmwareUpdateCoordinator` has a **documented known limitation**: during a connected-device auto-update, the device reboots into the bootloader, and Core's first-match enumeration (`WaitForBootloaderDeviceAsync` with no target path) could land on a *different* bootloader that the watcher is holding, if more than one is present. Location tracking would let the coordinator target the exact physical device it just rebooted, removing that limitation.

## Acceptance criteria (draft)

- [ ] A device retains a stable identity (its USB physical location) across serial → bootloader → post-flash transitions.
- [ ] With ≥2 identical bootloaders connected, the app shows and flashes the correct physical device.
[...]

PR #656, opened 2026-06-29
feat(firmware): app-global watcher holds & flashes all detected HID bootloaders

cptkoolbeenz:
## What

Replaces the dialog-scoped, single-device bootloader hold with an **app-global `BootloaderWatcher`** that discovers **every** sitting PIC32 HID bootloader and holds each one's USB handle open (its own exclusive transport, keyed by device path) with a continuously-pending keep-alive read — so none can be wedged by Windows USB selective-suspend (daqifi-nyquist-firmware#568) before the user flashes, **regardless of whether the Connection dialog is open**. The user can then flash any one of several connected bootloaders by path while the others stay held and wedge-proof.

> Supersedes #654 (single-device grab-and-hold). The watcher generalizes and replaces that approach — the dialog-scoped trigger was insufficient in the field — so this PR carries the full grab-and-hold → multi-device story and #654 is closed as superseded.

## Why

Multiple PIC32 bootloaders are **identical**: same `VID 0x04D8 / PID 0x003C`, and **no serial number** (`GetSerialNumber` throws). The only discriminator is the **USB device path**. Holding/flashing a specific one therefore requires path-based addressing, added in **`Daqifi.Core 0.27.0`** (`IHidTransport.ConnectByPathAsync` + a path-targeted `UpdateFirmwareAsync` overload; daqifi-core#275).

## How

- **`BootloaderWatcher`** (app-startup singleton): holds all detected bootloaders; `PrepareFlashAsync(path)` releases **only** the target (so the shared flasher reopens it by path) while every other bootloader stays held, then re-grabs it on a failed flash or drops it once a successful flash leaves it in application mode; `SuspendDiscoveryAsync()` lets the auto-update coordinator flash a device that reboots into the bootloader without the watcher stealing it.
- **`HidBootloaderDiscovery`**: continuous Core `HidDeviceFinder` loop (the exclusive holds make its per-cycle opens on already-held devices fail harmlessly).
- **`BootloaderHoldService`**: generalized to path addressing (`ConnectByPathAsync`) + a `HoldDropped` event on device removal; now owns and disposes its per-device transport.
- **Connection dialog** binds the Firmware tab to the watcher's list and flashes the chosen device by path; **firmware dialog** brackets the flash in the watcher lease; **`FirmwareUpdateCoordinator`** suspends discovery around the PIC32 flash.

## Code review

Ran a multi-agent review (4 dimensions × adversarial verification) of the watcher change before this PR: 5 confirmed findings (3 distinct), all fixed — per-device transport disposal, guarded discovery-resume across overlapping flash + auto-update, and a dialog `CollectionChanged` unsubscribe on a missed `Close()`. Four lower-confidence findings were verified and refuted.

## Tests

- New: 8 `BootloaderWatcher` tests + 5 `BootloaderHoldService` tests (path connect, `HoldDropped`, graceful release, transport disposal, overlapping flash/suspend resume coordination).
- **482 desktop unit tests green** (`TestCategory!=Ui`).

## Known limitation / follow-up

During a connected-device auto-update, if **another** bootloader is separately held, Core's first-match enumeration could land on the held one — auto-update is fundamentally a single-device operation, so this is documented and accepted. Tracked by #655 (track each device by physical USB location so identity survives serial ⇄ bootloader ⇄ post-flash and disambiguates identical bootloaders).

### DIFF
--- a/Daqifi.Desktop.DataModel/Daqifi.Desktop.DataModel.csproj
+++ b/Daqifi.Desktop.DataModel/Daqifi.Desktop.DataModel.csproj
@@ -8,7 +8,7 @@
 	</PropertyGroup>
 	<ItemGroup>
 	  <PackageReference Include="CommunityToolkit.Mvvm" Version="8.4.2" />
-        <PackageReference Include="Daqifi.Core" Version="0.25.0" />
+        <PackageReference Include="Daqifi.Core" Version="0.26.0" />
 	  <PackageReference Include="Roslynator.Analyzers" Version="4.15.0">
 		<PrivateAssets>all</PrivateAssets>
 		<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
--- a/Daqifi.Desktop.DataModel/Daqifi.Desktop.DataModel.csproj
+++ b/Daqifi.Desktop.DataModel/Daqifi.Desktop.DataModel.csproj
@@ -8,7 +8,8 @@
 	</PropertyGroup>
 	<ItemGroup>
 	  <PackageReference Include="CommunityToolkit.Mvvm" Version="8.4.2" />
-        <PackageReference Include="Daqifi.Core" Version="0.26.0" />
+        <!-- DEV: local Core 0.27.0 until 0.27.0 ships to NuGet (revert to PackageReference 0.27.0 for the PR). -->
+        <ProjectReference Include="..\..\daqifi-core\src\Daqifi.Core\Daqifi.Core.csproj" />
 	  <PackageReference Include="Roslynator.Analyzers" Version="4.15.0">
 		<PrivateAssets>all</PrivateAssets>
 		<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
--- a/Daqifi.Desktop.DataModel/Daqifi.Desktop.DataModel.csproj
+++ b/Daqifi.Desktop.DataModel/Daqifi.Desktop.DataModel.csproj
@@ -8,8 +8,7 @@
 	</PropertyGroup>
 	<ItemGroup>
 	  <PackageReference Include="CommunityToolkit.Mvvm" Version="8.4.2" />
-        <!-- DEV: local Core 0.27.0 until 0.27.0 ships to NuGet (revert to PackageReference 0.27.0 for the PR). -->
-        <ProjectReference Include="..\..\daqifi-core\src\Daqifi.Core\Daqifi.Core.csproj" />
+	  <PackageReference Include="Daqifi.Core" Version="0.27.0" />
 	  <PackageReference Include="Roslynator.Analyzers" Version="4.15.0">
 		<PrivateAssets>all</PrivateAssets>
 		<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
--- a/Daqifi.Desktop.IO/Daqifi.Desktop.IO.csproj
+++ b/Daqifi.Desktop.IO/Daqifi.Desktop.IO.csproj
@@ -7,8 +7,7 @@
 	</PropertyGroup>
 	<ItemGroup>
 		<PackageReference Include="System.IO.Ports" Version="10.0.9" />
-        <!-- DEV: local Core 0.27.0 until 0.27.0 ships to NuGet (revert to PackageReference 0.27.0 for the PR). -->
-        <ProjectReference Include="..\..\daqifi-core\src\Daqifi.Core\Daqifi.Core.csproj" />
+		<PackageReference Include="Daqifi.Core" Version="0.27.0" />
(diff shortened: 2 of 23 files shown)


cptkoolbeenz:
Thanks — addressed in `294333b`:

**Bugs**
- **UI deadlock via Invoke (#5):** `InvokeOnUiThread` now uses non-blocking `Dispatcher.BeginInvoke` instead of `Invoke`, so a bound-collection mutation under `_gate` can no longer lock-invert with `Dispose()`'s synchronous `_gate.Wait()`. Fixed.
- **Lease dispose can throw (#6):** `flashLease.DisposeAsync()` is now wrapped in try/catch (logged) so a watcher-resume failure can't crash the command or mask the flash outcome already recorded in the catch blocks. Fixed.
- **Discovery restart blocked (#7):** `HidBootloaderDiscovery.Stop()` now clears `_loopTask`, so a `Stop→Start` resume restarts discovery immediately rather than no-opping until the old loop drains. Fixed.

**Rule violations**
- **Two public types per file (#1, #2):** split `BootloaderDiscoveredEventArgs` and `BootloaderHoldDroppedEventArgs` into their own files. Fixed.
- **>120-char log line (#3):** wrapped. Fixed.
- **`new HidDeviceFinder()` (#4):** the Core finder is now created via an injected `Func<HidDeviceFinder>` factory (default = `new`), keeping the recreate-per-Start/Stop lifecycle while satisfying the inject-or-factory rule. Fixed.
- **Tab indentation in `.csproj` (#8):** **declined.** The `.csproj` is tab-indented throughout; the new `PackageReference` line matches the file's existing style, and the same finding was rejected in #551/#499. Changing only this line would make it inconsistent with the rest of the file.

482 unit tests green; will re-run the review.

cptkoolbeenz:
Pass 3 addressed in the latest commit: fixed the `BeginInvoke` shutdown edge that could strand a disposed hold in `_holds` (now non-throwing + ownership transferred before the UI marshal); exposed `Bootloaders` as a `ReadOnlyObservableCollection`; explicit `_discovery.Stop()` in `Dispose()`; validated `devicePath` in `BootloaderHoldDroppedEventArgs`; added ctor XML docs. The `.csproj` tab-indent finding remains declined (file is tab-indented throughout; same finding was rejected in #551/#499). 482 tests green.

cptkoolbeenz:
Pass 4 addressed: whitespace-validate `devicePath` in `HeldBootloader`; bounded `_gate.Wait()` (2s) in `BootloaderWatcher.Dispose` so a stuck gate-holder can't hang shutdown; null `_connectionDialogViewModel` after close. **Declined** the suggestion to drain the keep-alive under `_gate` in `BootloaderHoldService.Dispose` via an unbounded `GetAwaiter().GetResult()` — its own rationale notes it risks a shutdown deadlock if the HID driver is unresponsive, and it contradicts the bounded-wait hardening; the existing bounded 2s drain is safer. (`.csproj` tab still declined per file convention + #551/#499.) 482 tests green.

cptkoolbeenz:
Pass 5: fixed the post-dispose UI-mutation bug — the `BeginInvoke`-queued Add/Remove callbacks now check `_disposed` so a callback queued just before teardown can't re-add a phantom row (the Dispose `Clear` stays unguarded so final cleanup still runs). Declined the refactor suggestions (dispose-outside-gate, awaitable-cleanup, require-watcher-from-DI, unbounded keep-alive drain) and the `.csproj` tab, each with rationale in the commit. 482 tests green.

cptkoolbeenz:
Pass 6: fixed the dispose-timeout hold leak (Dispose now releases the holds — and their exclusive HID handles — best-effort whether or not the gate is acquired) and null the dialog VM field before Close() to avoid reentrancy. Declined the remaining refactor suggestions (async/awaitable disposal, dispose-outside-gate, require-watcher-from-DI, unbounded keep-alive drain), the XAML fallback (the model already guarantees a non-empty DisplayName), and the .csproj tab — rationale in the commit. 482 tests green.

cptkoolbeenz:
Pass 7: wrapped the one >120-char log line I'd just introduced. **Declining the remaining two and calling the cycle converged:**

- **"Dispose races _holds"** — this is a deliberate leak-vs-race tradeoff on the bounded-gate-timeout path only. It triggers solely when a handler is stuck holding the gate for >2s during shutdown (already pathological, e.g. a hung HID connect). The disposal snapshots the holds (no enumeration error) and `BootloaderHoldService.Dispose` is idempotent (guarded), so a concurrent dispose is safe; the alternative — disposing only under the gate — reintroduces the exclusive-HID-handle leak that pass 6 fixed. I'm choosing guaranteed resource release at shutdown over a benign theoretical race.
- **`.csproj` tab** — file is tab-indented throughout (#551/#499 precedent).

Summary across 7 passes: every genuine bug fixed; the recurring `Dispose`/UI-marshal findings were a self-correcting hardening cascade that's now bottomed out. 482 unit tests green, CI green. Treating this as converged.

PR #658, opened 2026-06-29
fix(build): stamp AssemblyVersion/FileVersion 3.3.0.0

tylerkron:
## What & why

`Directory.Build.props` hardcoded `AssemblyVersion`/`FileVersion` at `3.2.0.0` solution-wide. The 3.3.0 release bump (#657) updated only the csproj `<Version>` (→ informational `3.3.0+sha`) and the WiX `Product.wxs` (→ MSI ProductVersion `3.3.0.0`), but missed this file — so the shipped **3.3.0 build self-reported `3.2.0.0`**:

- The app **title bar reads "DAQiFi v3.2.0"** on the 3.3.0 build (confirmed on the bench).
- The exe `FileVersion` is `3.2.0.0`.
- **Sentry tags every 3.3.0 crash as `release: 3.2.0.0`** — `AppLogger` sets `options.Release = Assembly.GetName().Version`, so the stale assembly version corrupts the release tag we rely on to monitor post-release issues (notably the firmware-flash telemetry).

## Change

Bump `AssemblyVersion`/`FileVersion` to `3.3.0.0` to match the release. One line each.

After the fix, the built `DAQiFi.dll` reports:

```
AssemblyVersion = 3.3.0.0
FileVersion     = 3.3.0.0
ProductVersion  = 3.3.0+<sha>
```

The MSI `ProductVersion` (from `Product.wxs`) was already correct, so **install/upgrade behavior is unchanged** — this only corrects the exe's self-reported version + crash telemetry.

> Note: the existing 3.3.0 **draft** release MSI predates this commit and still self-reports 3.2.0.0. After merging, re-generate the release so the MSI carries 3.3.0.0.

## Verification

- `dotnet build -c Release` — succeeds, assembly version confirmed `3.3.0.0`.
- Full unit suite (483 tests) and the FlaUI hardware UI suite were green on this code prior to the change; this change is build-metadata only.

## Follow-up (optional, not in this PR)

Consider driving `AssemblyVersion`/`FileVersion` from the csproj `<Version>` (single source of truth) so future release bumps only touch one place.

### DIFF
--- a/Daqifi.Desktop/app.Debug.manifest
+++ b/Daqifi.Desktop/app.Debug.manifest
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
-  <assemblyIdentity version="3.2.0.0" name="MyApplication.app"/>
+  <assemblyIdentity version="3.3.0.0" name="MyApplication.app"/>
   <trustInfo xmlns="urn:schemas-microsoft-com:asm.v2">
     <security>
       <requestedPrivileges xmlns="urn:schemas-microsoft-com:asm.v3">
--- a/Daqifi.Desktop/app.manifest
+++ b/Daqifi.Desktop/app.manifest
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
-  <assemblyIdentity version="3.2.0.0" name="MyApplication.app"/>
+  <assemblyIdentity version="3.3.0.0" name="MyApplication.app"/>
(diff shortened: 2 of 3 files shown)


tylerkron:
**Re: Qodo "Stale manifest version" finding — ✅ Fixed in 6f90343.**

Good catch, and correct: bumping only `AssemblyVersion`/`FileVersion` left the embedded Windows manifests (`app.manifest`, `app.Debug.manifest`) still declaring `assemblyIdentity version="3.2.0.0"`, which is inconsistent with the shipped 3.3.0 build — and prior release bumps (#414, #503) did update these manifests, so this is the established convention.

Bumped both manifests' top-level `assemblyIdentity` to `3.3.0.0`. The `Microsoft.Windows.Common-Controls` `dependentAssembly` version (`6.0.0.0`) is unrelated and intentionally left unchanged. Verified with a clean `Release` build.

On the summary-level follow-up suggestion (derive these from a single version source / add a CI guard): agreed it's the right long-term direction, but out of scope for this minimal release fix — leaving it as a tracked follow-up.

Issue #664, opened 2026-07-03
feat: PWM output UI — expose Core's PWM support (per-channel duty, device-wide frequency)


tylerkron:
## Context

daqifi-core PR [daqifi/daqifi-core#281](https://github.com/daqifi/daqifi-core/pull/281) adds hardware-validated PWM output support (verified on a real Nyquist 1: streamed duty fractions matched commanded 30/50/80/90% through a wired loopback, frequency confirmed at 10.0 Hz by edge counting). Desktop should put a UI layer over it, alongside the digital-output UI (#663).

## What Core provides (after the next Daqifi.Core release)

- `IDigitalChannel.IsPwmCapable` — populated by Core from the firmware board mask. **Only channels 0, 3, 4, 5, 6, 7 support PWM** (identical across Nyquist variants). Also `IsPwmEnabled` and `PwmDutyCyclePercent` bookkeeping (preserved across status refreshes).
- `IStreamingDevice.SetPwmDutyCycle(channel, 1-100)` — per channel; live changes take effect immediately while enabled.
- `IStreamingDevice.SetPwmFrequency(6-50000 Hz)` — **device-wide**: all PWM channels share one hardware timer, so there is exactly one frequency for the whole device (this is why Core's API is device-level, not per-channel). `IStreamingDevice.PwmFrequencyHz` reports the last value commanded this session.
- `IStreamingDevice.SetPwmEnabled(channel, bool)` — start/stop. Core hard-blocks enabling on non-capable channels (a firmware bug otherwise freezes the channel's digital writes) and rejects duty 0 (firmware stores but never applies it — disable is the off switch).

Recommended call order (documented in Core): duty → frequency → enable.

## Suggested UI

- On PWM-capable digital channels: a PWM mode toggle plus a duty-cycle control (1–100 %). Hide/disable PWM controls on non-capable channels (`IsPwmCapable`).
- One **device-level** frequency control (6–50,000 Hz) — not per channel. Changing it while channels run is safe and rescales all enabled channels live.
- A channel in PWM mode should suppress its digital output controls (the hardware ignores direction/state writes while PWM runs; Core keeps the bookkeeping).
- After disabling PWM the pin is transiently high-impedance; the next digital state/direction write (or streaming tick) resumes driving from stored state — no special UI handling needed beyond reflecting `IsPwmEnabled = false`.

## Notes from hardware validation (core#281)

[...]

PR #669, opened 2026-07-06
chore(deps): bump Daqifi.Core 0.27.0 -> 1.0.0

tylerkron:
## Summary

Bumps `Daqifi.Core` from **0.27.0** to **1.0.0** in all three consuming projects (`Daqifi.Desktop`, `Daqifi.Desktop.DataModel`, `Daqifi.Desktop.IO`).

Core 1.0.0 is the stable-API milestone release ([release notes](https://github.com/daqifi/daqifi-core/releases/tag/v1.0.0)). It brings:

- **Decoded per-frame sample pipeline** (core#279) — additive API (`IChannel.SampleReceived`, `IDataSample.RawValue`/`DeviceTimestamp`); existing `MessageReceived` consumers unchanged, so desktop's stream decode keeps working as-is.
- **Digital output first-class** (core#280) — includes the by-channel-number stream decode fix desktop already relies on since #667.
- **PWM output API** (core#281) — `SetPwmEnabled` / `SetPwmDutyCycle` / `SetPwmFrequency` / `PwmFrequencyHz` + `IDigitalChannel.IsPwmCapable`/`IsPwmEnabled`/`PwmDutyCyclePercent`. This is the foundation for the PWM UI (#664), which follows in its own PR.

## Break points checked (from past bump experience)

| Known break point | This bump |
|---|---|
| `System.Management` version pin floor | No change needed — restore is clean with the existing 10.0.9 pin |
| `BootloaderSessionStreamingDeviceAdapter` (new `IStreamingDevice` members) | **Hit.** Core's `IStreamingDevice` gained the four PWM members; stubbed no-op like the other channel-management members (the adapter serves bootloader-only dialogs and never drives outputs) |
| `InitializeAsync` test fixture signature | No change needed — compiles and passes as-is |

Full compile surface: exactly 4 errors, all in the adapter; nothing else in desktop touches the changed Core surface.

## Testing

- **Fast gate** (`dotnet test --filter "TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests"`): **511 passed, 0 failed**.
- **FlaUI hardware suite** (`dotnet test Daqifi.Desktop.UITest`, USB Nyquist 1 attached): **17 passed, 2 failed** — the two failures are the known-environmental SD-card read/import scenarios on this bench (firmware SD-read bug: import of a freshly written file hangs / parses to no samples; fails identically on `main`, tracked in firmware#567 territory). All connect / stream / plot / export / profiles / digital-output / disconnect-reconnect scenarios green.


tylerkron:
Replies to the Qodo review findings (both are in the summary comment only, so responding top-level):

**1. PWM members missing XML docs (rule 244813)** — *partially addressed.* This adapter deliberately documents its stubbed interface surface with one shared rationale comment over the whole no-op block rather than per-member `///` docs — the same convention this class has kept through the 0.24.0 member additions, and the same finding was reviewed and not adopted in #633 and #400. What *was* stale is that the shared comment didn't cover the new members: 926f4f2 extends it to name the 1.0.0 PWM members and to note that `PwmFrequencyHz => 0` matches Core's documented "none commanded this session" sentinel.

**2. PWM methods silently no-op** — *disagree, consistent with precedent (#633).* Silent no-op is this adapter's designed contract, not an accident: it exists solely to satisfy `IStreamingDevice` for bootloader-only update dialogs where the device is already in firmware mode — `Send()`, streaming, channel management, and DIO members are all deliberate no-ops for the same reason. Throwing `NotSupportedException` would turn an impossible-by-construction path (nothing in the firmware-update flow drives outputs) into a new crash surface inside firmware flashing, which is exactly where we least want one. `PwmFrequencyHz => 0` is not an invalid value — 0 is Core's own documented sentinel for "no frequency commanded this session".

PR #691, opened 2026-07-14
chore: drop vestigial Channel EF entity and its table

### LINKED ISSUE #678: chore: drop vestigial Channel EF entity and its table
## Change
`Daqifi.Desktop/Channel/Channel.cs` (71 lines) is a plain POCO `Channel : IChannel` that — unlike `AnalogChannel`/`DigitalChannel` — wraps no Core channel and is never constructed anywhere in the solution. It is mapped as an EF entity (`entity.ToTable("Channel")` in `LoggingContext.OnModelCreating`) but has **no `DbSet<Channel>`**; the only references are EF migration Designer snapshots.

Delete the class, remove the mapping, and add a migration dropping the `Channel` table.

## Caveats
- Confirm nothing reads the `Channel` table at runtime (raw SQL, `SessionDataRepository`, importers) before dropping — existing user databases created by older versions will contain the table.
- The drop migration must be safe on databases where the table holds rows.

## Dependency
None — doable today.

_From a cross-repo audit to make daqifi-desktop a thin wrapper over Daqifi.Core (2026-07-11, Core 1.0.0)._

tylerkron:
Closes #678

## Summary
- Deletes `Daqifi.Desktop/Channel/Channel.cs` — a plain POCO `Channel : IChannel` that wraps no Core channel, is never constructed anywhere in the solution, and had no `DbSet<Channel>`.
- Removes its EF mapping (`modelBuilder.Entity<Channel.Channel>` / `ToTable("Channel")`) from `LoggingContext`.
- Removes the unused `LoggingSession.Channels` navigation property (the exporter derives channels from `DataSamples`, not this collection — confirmed via `LoggingSessionSampleSource.GetChannels()`).
- Adds EF migration `DropChannelTable` that drops the `Channel` table, safe for existing user databases that still have the table (and rows) from older app versions.
- Cleans up `OptimizedLoggingSessionExporterTests`, which built `Channel.Channel` instances and assigned them to `LoggingSession.Channels` but the exporter never read that collection — dead test setup.

## Caveats addressed
- Confirmed no raw SQL, `SessionDataRepository`, or importer reads the `Channel` table — the only references besides the mapping were EF migration Designer snapshots.
- The generated migration also fixed pre-existing drift between the model snapshot and the actual `Samples` index name (`IX_Samples_LoggingSessionID_TimestampTicks` → `IX_Samples_SessionTime`), which `dotnet ef migrations add` picked up as part of diffing the full model — this was already-pending model drift the tests suppress via `Ignore(PendingModelChangesWarning)`, now resolved.
- Per Qodo review feedback, `DropChannelTable`'s `Up()` uses raw SQL with `IF EXISTS` guards so the drop doesn't fail on user databases that were hand-modified or already partially migrated.

## Test plan
- [x] `dotnet build "DAQiFi Desktop.sln"` — 0 errors
- [x] `dotnet test --filter "TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests"` — 516 passed, 0 failed
- [x] Verified no runtime code paths read the `Channel` table before dropping it
- [x] Verified the `IF EXISTS`-guarded migration against a real SQLite DB via the `Migrate()`-based test suite


tylerkron:
Thanks for the review — addressed the high-level suggestion in 83d5bef: hardened the `DropChannelTable` migration's `Up()` to use raw SQL with `IF EXISTS` guards (`DROP TABLE IF EXISTS "Channel"` / `DROP INDEX IF EXISTS "IX_Samples_LoggingSessionID"`) instead of the strongly-typed `DropTable`/`DropIndex` operations, so the upgrade won't fail on user databases that were hand-modified or already partially migrated. Verified against a real SQLite DB via the existing `Migrate()`-based test suite (39 tests exercising real migrations, plus the full 516-test fast gate) — all green.

Left `Down()` and the `RenameIndex` call as EF-generated, since those are dev-only rollback paths, not something that runs against live user data on upgrade.

PR #697, opened 2026-07-15
chore: remove redundant serial GetDeviceInfo poll loop

### LINKED ISSUE #618: refactor: Remove serial GetDeviceInfo poll loop once Core init blocks until populated
## Change
Remove the serial `GetDeviceInfo` re-request poll loop (every 1s, 8s timeout) in `SerialStreamingDevice` (`Daqifi.Desktop/Device/SerialDevice/SerialStreamingDevice.cs:135-175`) once Core's serial `InitializeAsync`/`ConnectSerialAsync` blocks until `ChannelsPopulated` (or fails). The desktop should then receive a populated device directly from Core.

## Dependency
**Blocked by daqifi-core#247** (serial init blocks until channels populate).

_From a cross-repo audit to make daqifi-desktop a thin wrapper over Daqifi.Core._

tylerkron:
## Summary
- `SerialStreamingDevice.OnCoreDeviceInitialized` re-implemented an 8s-timeout/1s-resend `GetDeviceInfo` poll loop to wait for the device to populate its channels.
- Since Core 1.0.0 (daqifi-core#247, consumed since #669), `AbstractStreamingDevice.Connect()` already blocks on `coreDevice.InitializeAsync()` before calling `OnCoreDeviceInitialized()` — and `InitializeAsync` internally does the exact same thing: sends `GetDeviceInfo`, waits for `ChannelsPopulated`, re-sends every 1s, and throws `TimeoutException` after 8s (identical defaults to the loop being removed).
- By the time `OnCoreDeviceInitialized` ran, initialization had already succeeded or `Connect()` had already failed via the exception path — the serial-specific loop was pure dead-weight duplication.
- Removed `OnCoreDeviceInitialized`, `OnCoreChannelsPopulated` overrides, the `_initialStatusReceivedSource` completion source, and the three timeout/interval constants. Serial now falls through to the base (no-op) `OnCoreDeviceInitialized`, matching the WiFi device's existing pattern.
- `LogConnectFailure`'s `TimeoutException` case is untouched — Core's `InitializeAsync` throws the same exception type with an equivalent message, so the existing warning-level classification (issue #517, #632) still applies correctly.

Closes #618

## Test plan
- [x] `dotnet test --filter "TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests"` — 545 passed, 0 failed
- [x] `dotnet test Daqifi.Desktop.UITest --filter "FullyQualifiedName~ManualSerialConnectTests"` against a physically attached DAQiFi device on COM3 — all 3 tests passed, including `ManualSerial_RealDevicePort_Connects` which drives a real end-to-end serial connect through the new code path

### DIFF
--- a/Daqifi.Desktop/Device/SerialDevice/SerialStreamingDevice.cs
+++ b/Daqifi.Desktop/Device/SerialDevice/SerialStreamingDevice.cs
@@ -11,14 +11,9 @@ namespace Daqifi.Desktop.Device.SerialDevice;
 
 public class SerialStreamingDevice : AbstractStreamingDevice, ILanChipInfoProvider
 {
-    private static readonly TimeSpan InitialStatusTimeout = TimeSpan.FromSeconds(8);
-    private static readonly TimeSpan InitialStatusPollInterval = TimeSpan.FromMilliseconds(100);
-    private static readonly TimeSpan InitialStatusRequestInterval = TimeSpan.FromSeconds(1);
-
     #region Properties
     private SerialPort? _port;
     private SerialStreamTransport? _transport;
-    private TaskCompletionSource<bool>? _initialStatusReceivedSource;
 
     public SerialPort? Port
     {
@@ -89,9 +84,6 @@ public class SerialStreamingDevice : AbstractStreamingDevice, ILanChipInfoProvid
     /// </summary>
     protected override CoreStreamingDevice CreateCoreDevice()
     {
-        _initialStatusReceivedSource = new TaskCompletionSource<bool>(
-            TaskCreationOptions.RunContinuationsAsynchronously);
-
         // Use Core's transport for unified message handling (both send and receive).
         // The transport manages the actual SerialPort connection internally; DTR must stay
         // enabled or the device will not stream over USB.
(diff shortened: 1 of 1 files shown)


tylerkron:
Re: Qodo's "Connect returns before sync" finding — this isn't a real race.

`AbstractStreamingDevice.Connect()` subscribes `coreDevice.ChannelsPopulated += OnCoreChannelsPopulated` (the desktop hydration handler, added in #404) *before* it calls `coreDevice.InitializeAsync()`. `InitializeAsync()` only sends `GetDeviceInfo` and subscribes its own internal completion handler to the same event *after* that point (inside `WaitForChannelsPopulatedAsync`). `ChannelsPopulated` is a plain synchronous multicast delegate (`ChannelsPopulated?.Invoke(...)` in `DaqifiDevice.cs`), and C# invokes multicast subscribers in subscription order — so the desktop's hydration handler always runs, synchronously, before Core's internal wait is satisfied and `InitializeAsync()` is allowed to return. There's no window where `Connect()` can return with `ChannelsPopulated` fired but the desktop sync not yet run.

This is the same concern Qodo's own "Relevance" note flags as previously raised and rejected on #404, for the same reason. No code change made.

PR #710, opened 2026-07-16
chore: delete dead code (excluded VMs, unused converters, dead style, unused images, redundant package refs)

### LINKED ISSUE #703: chore: delete dead code (excluded VMs, unused converters, dead style, unused images, redundant package refs)
## Change

Straight dead-code deletion, no Core dependency and no behavior change. ~460 lines, 9 images, 2 package references.

### 1. Files excluded from compilation (276 lines)

All three exist on disk and are referenced *only* by the `<Compile Remove>` lines that exclude them. They are not compiled, so nothing can reference them.

| File | Lines |
|---|---|
| `Daqifi.Desktop/ViewModels/PlottingViewModel.cs` | 178 |
| `Daqifi.Desktop/ViewModels/PlotViewModel.cs` | 68 |
| `Daqifi.Desktop/ViewModels/ViewModelLocator.cs` | 30 |

Delete the files **and** the whole `<ItemGroup>` at `Daqifi.Desktop/Daqifi.Desktop.csproj:30-34`.

### 2. Dead types (163 lines)

Each has exactly one whole-word hit repo-wide (its own declaration), checked against `*.cs` **and** `*.xaml` across all four projects, excluding bin/obj.

| File | Lines | Note |
|---|---|---|
| `Daqifi.Desktop/Models/AddProfileModel.cs` | 50 | |
| `Daqifi.Desktop/Converters/ConnectionTypeToUsbConverter.cs` | 21 | never registered in any XAML |
| `Daqifi.Desktop/Converters/BoolToStatusColorConverter.cs` | 21 | registered in `App.xaml:50`, zero `StaticResource` uses - delete file + App.xaml line |
| `Daqifi.Desktop/Converters/BoolToConnectionStatusConverter.cs` | 20 | never registered in any XAML |
| `Daqifi.Desktop/Converters/BoolToActiveStatusConverter.cs` | 20 | registered in `App.xaml:49`, zero uses - delete file + App.xaml line |
| `Daqifi.Desktop/Helpers/MyMultiValueConverter.cs` | 17 | |
| `Daqifi.Desktop.Common/Loggers/NoOpLogger.cs` | 14 | never constructed |

### 3. Dead XAML style (~21 lines)

`Daqifi.Desktop/App.xaml:20-40` - `<Style x:Key="FadeButton">`. The only other hit is a *comment* at `View/Prototype/LoggedDataPanePrototype.xaml:72` saying it replaced FadeButton.

### 4. Unused images (9 files, ~15 KB)

In `Daqifi.Desktop/Images/`: `1Nq.png`, `2Nq.png`, `3Nq.png`, `Add.png`, `Configuration.png`, `Icon_Large.png`, `Icon_Medium.png`, `Icon_Small.png`, `WiFi.png`.

A repo-wide `grep -ril` per filename (excluding bin/obj/.git/Images) returns only `Daqifi.Desktop.csproj`. No `pack://` URIs point at local images. The only real image references anywhere are `App.xaml.cs:209` (`SplashScreen("Images/DAQiFi.png")`) and `View/MigrationStatusWindow.xaml:26`. Delete the files plus their `<None Remove>` / `<Resource Include>` entries.

`WiFi.png` is doubly dead - it has a `<None Remove>` entry but no `<Resource Include>`, so it isn't even embedded today.

**Keep** `DAQiFi.png` (splash + migration window) and `WiFiDAQ.ico` (`<ApplicationIcon>`).

### 5. Redundant PackageReferences

`System.Configuration.ConfigurationManager` in both `Daqifi.Desktop/Daqifi.Desktop.csproj` and `Daqifi.Desktop.Test/Daqifi.Desktop.Test.csproj`. `grep -rn "ConfigurationManager|using System.Configuration" --include=*.cs` returns zero hits in either project. The only real use is `Daqifi.Desktop.Common/Loggers/AppLogger.cs`, and Common declares its own reference.

## Checked and explicitly NOT dead

Recording these so the next audit doesn't re-litigate them:

- `Daqifi.Desktop/Loggers/LoggingContextDesignTimeFactory.cs` - zero code references, but EF Core tooling instantiates it by reflection via `IDesignTimeDbContextFactory<LoggingContext>`. **Do not delete.**
- `Resources/DarkDialog.xaml` - not merged in App.xaml, but merged per-dialog by `DebugWindow.xaml:19`, `DuplicateDeviceDialog.xaml:14`, `ExportDialog.xaml:16`, `FirmwareDialog.xaml:17`.
- `InvertedBoolToVisibilityConverter` - registered *locally* at `DeviceLogsView.xaml:16` and used at line 212, shadowing the App.xaml key of the same name.
- `ConnectionTypeToColorConverter`, `BrushColorMatchConverter`, `OxyColorToBrushConverter`, `EnumDescriptionConverter`, `BoolAndToVisibilityConverter` - all used via `StaticResource`.
- `SQLitePCLRaw.bundle_e_sqlite3` - no code references, but it's the runtime native provider for Microsoft.Data.Sqlite.
- `Microsoft.Xaml.Behaviors.Wpf` - used via `xmlns:i=` in MainWindow.xaml and two Prototype XAMLs.

Also confirmed clean: no `.proto` files anywhere and no local `DaqifiOutMessage` (desktop consumes Core's type); no `#if false` blocks and no meaningful commented-out code.

_From a cross-repo audit to make daqifi-desktop a thin wrapper over Daqifi.Core (2026-07-16, Core 1.1.1)._

tylerkron:
## Summary

Straight dead-code deletion per #703, no behavior change:

- Deleted 3 ViewModels excluded from compilation (`PlottingViewModel.cs`, `PlotViewModel.cs`, `ViewModelLocator.cs`) and their `<Compile Remove>` entries in the csproj.
- Deleted 6 dead converter/helper types with zero references: `ConnectionTypeToUsbConverter`, `BoolToStatusColorConverter`, `BoolToConnectionStatusConverter`, `BoolToActiveStatusConverter`, `MyMultiValueConverter`, `NoOpLogger`. Removed the two `App.xaml` registrations for the converters that were still wired up but unused.
- Removed the dead `FadeButton` style from `App.xaml`.
- Deleted 9 unused images from `Daqifi.Desktop/Images/` plus their csproj `<None Remove>`/`<Resource Include>` entries.
- Removed the redundant `System.Configuration.ConfigurationManager` PackageReference from `Daqifi.Desktop.csproj` and `Daqifi.Desktop.Test.csproj` (Common already references it; no code in either project uses it directly).
- `AddProfileModel.cs` also declared the still-used `Profile`, `ProfileDevice`, and `ProfileChannel` types (referenced from `LoggingManager`, `ProfilesPaneViewModel`, `DaqifiViewModel`) — only the unused `AddProfileModel` class was removed, and the file was renamed to `ProfileModels.cs` to match its actual contents.

Closes #703

## Test plan
- [x] `dotnet build` — 0 errors
- [x] `dotnet test --filter "TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests"` — 550 passed, 0 failed


tylerkron:
Thanks @qodo-code-review — no action-required findings on this one, and no inline threads to reply to. Responding to the one recommendation in the high-level assessment:

**"Automate dead-code detection in CI"** — agreed in principle, and out of scope for this PR (which is a straight deletion with no behavior change). Worth noting the two halves have very different cost/benefit here:

- *Unused `PackageReference` detection* is tractable and I think worth doing. There's a real precedent for the failure mode this PR fixed — `System.Configuration.ConfigurationManager` was declared in two projects that never used it directly.
- *Unused XAML resource detection* is the harder half and where I'd be cautious about noise. This codebase has at least two patterns that a naive "declared but no `StaticResource` hit" check would flag as false positives: `Resources/DarkDialog.xaml` is merged per-dialog rather than in `App.xaml`, and `InvertedBoolToVisibilityConverter` is registered locally in `DeviceLogsView.xaml` shadowing an `App.xaml` key of the same name. A check that flags those every build gets muted fast.

I'd rather not bolt this on as an unscoped CI addition inside a deletion PR. Filing it as a follow-up issue instead so the analyzer choice and false-positive tuning get their own review.

One correction to the AI Description, for anyone reading this later: it says the test project now "relies on the dependency already being provided transitively." That's true but incidental — the point is that **no code in `Daqifi.Desktop.Test` or `Daqifi.Desktop` references `ConfigurationManager` at all**. The only real consumer is `Daqifi.Desktop.Common/Loggers/AppLogger.cs`, and Common declares its own reference. The removal isn't leaning on transitive resolution to keep a live usage compiling.

**"Consolidate XAML resource registration into feature dictionaries"** — noted, agreed it's a larger refactor than this PR's goal, not taking it here.

tylerkron:
Filed the dead-code CI check as #711.

tylerkron:
## Runtime verification (updating my earlier note that this was compile-time-only)

Deleting XAML resources is a runtime failure a green build misses, so I drove the real app (worktree Debug build, not the installed one) rather than relying on the build + unit tests.

**Result: launches clean, all panes and the connection dialog render, zero errors in the log.**

| Surface | Result |
|---|---|
| Startup | Splash -> `DAQiFi v3.3.0` main window, responding |
| Live Graph / Logged Data / Channels / Devices / Profiles | all 5 panes render |
| `ConnectionDialog` (merges `DarkDialog.xaml`) | opens, all 4 tabs render, device discovered |
| App log for the whole session | 2 lines, both INFO, **0 errors / 0 warnings / 0 XamlParseException** |

The Profiles pane matters most here since `ProfileModels.cs` was the only non-mechanical edit — it renders its empty state, and `DAQifiProfilesConfiguration.xml` was written during startup. `Logged Data` exercises `DeviceLogsView.xaml`, which carries the locally-registered `InvertedBoolToVisibilityConverter` that #703 flagged as must-keep; it resolves fine. A device was discovered over both WiFi (192.168.1.30) and USB (COM3).

### Two things I checked that the diff review can't settle on its own

**1. My original grep had a gap.** I verified the deleted converters by *type* name (`BoolToActiveStatusConverter`), but XAML consumes them by *resource key* (`BoolToActiveStatus`) — a shorter string that a grep for the longer type name cannot match. Re-grepped all `*.xaml` for the keys themselves (`BoolToActiveStatus`, `BoolToStatusColor`, `FadeButton`): zero usages, the only `FadeButton` hit being the comment at `LoggedDataPanePrototype.xaml:72` noting it was replaced. There are also no `FindResource`/`TryFindResource` calls anywhere, so there is no dynamic-key lookup that static analysis could miss. Conclusion held, but it wasn't actually proven before.

**2. Why dropping `System.Configuration.ConfigurationManager` is safe.** `AppLogger.cs:85` calls `ConfigurationManager.AppSettings["SentryDsn"]` at startup, and the DLL is *not* copied to the output dir — which looks alarming. It resolves because `Microsoft.WindowsDesktop.App` (10.0.8) ships `System.Configuration.ConfigurationManager.dll` in the shared framework. That's exactly why the split is correct:

- `Daqifi.Desktop` + `Daqifi.Desktop.Test` are `net10.0-windows` -> get it from the WindowsDesktop shared framework -> explicit refs were redundant.
- `Daqifi.Desktop.Common` is plain `net10.0` with no desktop framework -> genuinely needs its own reference -> kept.

Also worth noting there's no version-downgrade risk: Common pins the same `10.0.9` the removed refs did. Proven at runtime rather than by inspection — the "SentryDsn not found" warning did *not* fire this run, meaning `ConfigurationManager` successfully read a real 95-char DSN from `DAQiFi.dll.config`.

### Installer

`Product.wxs` harvests `$(var.SourceDir)\**` and takes its shortcut icon from the exe itself (`<ApplicationIcon>Images\WiFiDAQ.ico</ApplicationIcon>`, kept). The 9 deleted images were WPF-embedded `Resource`s, never loose files, and the installer's own `WiFiDAQ_Banner.bmp` / `WiFiDAQ_Dialog.bmp` live in the Setup project's separate `Images/` dir. Unaffected — worth stating explicitly since the installer builds from its own solution that this PR's CI never compiles.

PR #721, opened 2026-07-19
chore: add daqifi-loop autonomous work-loop skill

tylerkron:
## What

Adds a project-scoped Claude Code skill at `.claude/skills/daqifi-loop/SKILL.md` that stands up the autonomous work loop for this repo — the same loop already used on daqifi-core, adapted for desktop.

## Why here

Checking it into the repo means it travels with a clone/pull — including onto a Windows machine — so the loop can be started against the desktop app with `/daqifi-loop` in a Claude Code session on that box.

## What the loop does

A recurring cron that, each fire, spawns one subagent to do ONE unit of work:
1. shepherd open PRs through Qodo/CI,
2. work the next eligible backlog ticket,
3. when the defined backlog is dry, do a self-directed improvement (research → file a tracked issue → work it),

up to a concurrency cap of unreviewed PRs. **It never merges** — the user reviews/merges. Session-only (dies with the session, 7-day auto-expiry).

## Desktop-specific adaptations from the core version

- **Self-locating** (`git rev-parse`) and **auto-discovers the board** each fire — no hardcoded path or COM port, so the file is machine-agnostic.
- **Windows-aware:** no Homebrew PATH step; target frameworks derived from the csproj rather than assuming net9+net10.
- **Bench honesty:** WPF has no clean headless drive, so the full `dotnet test` suite is the validation gate; the loop won't script the GUI or treat 'couldn't bench the UI' as a failure.
- **Destructive-op denylist** hardened for Windows (HID/firmware flash paths work here, unlike macOS — stay read/stream-only).
- Structured as a **generic body + single REPO PROFILE block** so it doubles as the template for retargeting other .NET DAQiFi repos.

Not urgent; for review — no code paths touched, docs/tooling only.


tylerkron:
Thanks — verified all four against `CLAUDE.md` and the repo layout. Fixed the substantive three in 5f43250:

**2. Wrong test gate — agreed, real bug.** `CLAUDE.md` documents the fast unit gate (`dotnet test --filter "TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests"`) vs. the on-demand `Daqifi.Desktop.UITest` (FlaUI) integration gate. The skill now mandates the unit gate every fire and makes UITest an explicit optional/on-demand step (interactive Windows session + attached device), logging-and-proceeding when prerequisites aren't met.

**3. Incorrect macOS compatibility — agreed.** Narrowed the claim: the *file* is machine-agnostic, but the `net*-windows` test projects only execute on Windows, so the loop must run on Windows (macOS = build-only / Windows CI). Added an explicit OS note.

**4. SESSION_LOG not bootstrapped — agreed.** The task now creates `SESSION_LOG.md` with a header if missing before the first read/append.

**1. Lines >120 — partially agree, not mechanically fixing.** Two of the flagged ranges can't be wrapped without breaking the file: the frontmatter `description` is a single YAML scalar, and the `TASK` blockquote is passed **verbatim** as a subagent prompt — hard-wrapping it would inject breaks into the prompt text. This is a prompt-document, not source, so a code line-length rule is a poor fit here. Happy to add a `.markdownlint`/lint-ignore for this path if the repo wants the rule to formally not apply.

PR #724, opened 2026-07-20
fix: classify serial transport-closed mid-init as environmental (#588)

### LINKED ISSUE #588: System.InvalidOperationException: The BaseStream is only available when the port is open.
Sentry Issue: [DAQIFI-DESKTOP-X](https://daqifi.sentry.io/issues/7544619659/)

```
System.InvalidOperationException: The BaseStream is only available when the port is open.
  File "SerialStreamTransport.cs", line 53, in Stream SerialStreamTransport.get_Stream()
    return _serialPort?.BaseStream ?? throw new InvalidOperationException("Transport is not connected.");
  ?, in async Task<IReadOnlyList<string>> DaqifiDevice.ExecuteTextCommandCoreAsync(Func<CancellationToken, Task> setupActionAsync, int responseTimeoutMs, int completionTimeoutMs, CancellationToken cancellationToken)
  File "DaqifiStreamingDevice.cs", line 131, in async Task DaqifiStreamingDevice.InitializeAsync()
    var lines = await ExecuteTextCommandAsync(
  File "Device\SerialDevice\SerialStreamingDevice.cs", line 112, in bool SerialStreamingDevice.Connect()
```
This issue was automatically created by Sentry via [GitHub Alert](https://sentry.io/organizations/daqifi/monitors/alerts/3262760/)

tylerkron:
## Problem

Closes #588.

`SerialStreamingDevice` opens the COM port in `CreateCoreDevice`, then Core's `InitializeAsync` reads/writes it. If the port closes in that window — the device is unplugged, powered off, or the USB driver drops the port mid-connect — accessing the transport stream throws an `InvalidOperationException`:

- .NET's `SerialPort.BaseStream` getter: *"The BaseStream is only available when the port is open."* (the exact message in the Sentry issue)
- Core's `SerialStreamTransport`: *"Transport is not connected."*

Both fell through `LogConnectFailure`'s `default` branch, which calls `AppLogger.Error` and captures the exception to Sentry as an app bug — even though it's the same kind of user/environmental condition already downgraded for connect-timeout (#517/#632) and SCPI-init (#589).

## Fix

- Add an `InvalidOperationException when IsTransportClosedError(ex)` case to `SerialStreamingDevice.LogConnectFailure` that logs a Warning (no Sentry capture) instead of an Error.
- `IsTransportClosedError` is a pure, `internal static` predicate matching the two specific phrases (not the bare word "port"), so unrelated `InvalidOperationException` bugs still hit the default Error path — mirroring the existing `IsScpiInitializationError` design.

## Why it's safe

`CreateCoreDevice` always calls `_transport.Connect()` (which opens the port) before init runs; a failure to *open* throws earlier with a different message and is unaffected. Reaching the stream-getter-throws state therefore means the port was open and then closed = environmental, so downgrading it does not mask an app logic bug.

## Tests

Added 5 unit tests to `SerialStreamingDeviceLogConnectFailureTests` (predicate matches both messages, is case-insensitive, rejects an unrelated message, and `LogConnectFailure` does not throw for the transport-closed case). Full unit gate green: **553 passed, 0 failed**.

Not merging — opened for review.

### DIFF
--- a/Daqifi.Desktop.Test/Device/SerialStreamingDeviceLogConnectFailureTests.cs
+++ b/Daqifi.Desktop.Test/Device/SerialStreamingDeviceLogConnectFailureTests.cs
@@ -3,10 +3,11 @@ using Daqifi.Desktop.Device.SerialDevice;
 namespace Daqifi.Desktop.Test.Device;
 
 /// <summary>
-/// Tests for <see cref="SerialStreamingDevice"/>'s connect-failure classification (issue #589):
-/// Core's SCPI-error-during-initialization InvalidOperationException is a device/environmental
-/// condition, not an app bug, so it must be downgraded to a Warning (no Sentry capture) rather
-/// than the default Error path.
+/// Tests for <see cref="SerialStreamingDevice"/>'s connect-failure classification. Certain
+/// InvalidOperationExceptions raised during connect are device/environmental conditions, not app
+/// bugs, so they must be downgraded to a Warning (no Sentry capture) rather than the default Error
+/// path: Core's SCPI-error-during-initialization message (issue #589) and the serial transport
+/// reporting the COM port closed mid-initialization (issue #588).
 /// </summary>
 [TestClass]
 public class SerialStreamingDeviceLogConnectFailureTests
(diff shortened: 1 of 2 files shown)


tylerkron:
Thanks Qodo. Declining this one — it's a summary-only, low-relevance recommendation and, as the finding itself notes, "Team prefers testing pure predicates, not logger side effects."

The classification behavior this finding worries about (Warning-vs-Error regression) IS guarded, via the pure predicate tests already in this PR: `IsTransportClosedError_MatchesDotNetBaseStreamMessage`, `IsTransportClosedError_MatchesCoreTransportNotConnectedMessage`, `IsTransportClosedError_IsCaseInsensitive`, and the regression guard `IsTransportClosedError_DoesNotMatchUnrelatedInvalidOperationException` (which asserts an unrelated InvalidOperationException stays on the default Error path). If someone regresses the predicate or the switch arm, those assertions fail.

The flagged `LogConnectFailure_*_DoesNotThrow` tests are deliberately non-asserting smoke tests, matching the established convention (PR #709/#720). As the finding acknowledges, `AppLogger` disables real logging in test mode, so asserting log level via logger side effects there isn't possible by design. No change.

tylerkron:
🔬 **Bench-tested on Windows against hardware** (smoke).

Launches clean and a live USB connect to the board succeeds — the connect subsystem that hosts the modified `LogConnectFailure` has no happy-path regression. The Error→Warning downgrade only fires on the specific mid-init transport-closed failure, which isn't GUI-forceable; that path stays covered by the unit tests in this PR. ✅

PR #745, opened 2026-07-20
fix: harden EnumDescriptionConverter against NRE on undefined enum values and non-Description attributes

### LINKED ISSUE #744: fix: EnumDescriptionConverter throws NullReferenceException on undefined enum values and non-Description attributes
## Problem

`EnumDescriptionConverter.GetEnumDescription` (Daqifi.Desktop/Helpers/EnumDescriptionConverter.cs) has two latent `NullReferenceException` paths in a converter that is registered app-wide (`App.xaml` key `EnumDescription`) and bound to device-reported `WifiMode`/`WifiSecurityType` values in `DevicesPanePrototype.xaml`.

### Bug 1 — undefined enum value (line 11-13)
```csharp
var fieldInfo = enumObj.GetType().GetField(enumObj.ToString());
var attribArray = fieldInfo.GetCustomAttributes(false); // NRE when fieldInfo == null
```
`Type.GetField(name)` returns `null` when the value is not a *named* member (an out-of-range/undefined value whose `ToString()` is the numeric string, e.g. `"999"`). Dereferencing `fieldInfo` then throws.

### Bug 2 — first attribute is not a DescriptionAttribute (line 20-21)
```csharp
var attrib = attribArray[0] as DescriptionAttribute;
return attrib.Description; // NRE when attribArray[0] is not a DescriptionAttribute
```
The code assumes the *first* custom attribute is a `DescriptionAttribute`. If the member carries any other attribute (or attribute order puts a non-Description attribute first), the `as` cast yields `null` and the `.Description` access throws. `GetCustomAttributes` order is not guaranteed.

## Proposed fix (non-breaking)
Null-guard `fieldInfo` (fall back to the value's string form), and select the `DescriptionAttribute` specifically via the typed `GetCustomAttributes(typeof(DescriptionAttribute), false)` + `FirstOrDefault`, falling back to the member name when absent. This preserves every current behavior (described member -> description, empty description -> empty string, no attribute -> name) while eliminating both crashes.

## Why it's safe
- Only changes behavior for inputs that currently *throw*; all existing `EnumDescriptionConverterTests` expectations are preserved.
- Fully unit-testable: add cases for an undefined value, a member with a non-Description attribute, and a member mixing Description + another attribute.

Found during autonomous codebase review.

tylerkron:
## Summary

`EnumDescriptionConverter.GetEnumDescription` (Daqifi.Desktop/Helpers/EnumDescriptionConverter.cs) had two latent `NullReferenceException` paths. The converter is registered app-wide (`App.xaml` key `EnumDescription`) and bound to device-reported `WifiMode`/`WifiSecurityType` values in `DevicesPanePrototype.xaml`, so a device (or any caller) supplying an out-of-range or oddly-attributed enum value could crash item rendering.

**Bug 1 — undefined enum value.** `Type.GetField(enumObj.ToString())` returns `null` when the value is not a named member (its `ToString()` is the numeric string, e.g. `"999"`), so dereferencing `fieldInfo` threw. Now falls back to the value's string form.

**Bug 2 — first attribute is not a DescriptionAttribute.** `attribArray[0] as DescriptionAttribute` assumed the first custom attribute is a `DescriptionAttribute`; any other attribute (or non-guaranteed attribute order) made the cast `null` and `.Description` threw. Now selects the `DescriptionAttribute` specifically via the typed `GetCustomAttributes` overload + `FirstOrDefault`, falling back to the member name when absent.

## Why it's safe / non-breaking

Behavior is preserved for every input that previously worked (described member -> description, empty description -> empty string, no attribute -> name). Only inputs that previously **threw** change behavior. Verified by the existing 11 tests plus 3 new ones (undefined value, member with only a non-Description attribute, member mixing Description + another attribute).

## Testing

- Unit gate (`dotnet test --filter "TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests"`): 643 passed / 0 failed (+3 new).
- App-launch smoke (`Launch_MainWindowAppears_AndIsResponsive`): passed.

Closes #744

_Opened by the autonomous work loop — not merging; for maintainer review._

### DIFF
--- a/Daqifi.Desktop.Test/Converters/EnumDescriptionConverterTests.cs
+++ b/Daqifi.Desktop.Test/Converters/EnumDescriptionConverterTests.cs
@@ -33,6 +33,20 @@ public class EnumDescriptionConverterTests
         Value3
     }
 
+    // Test enum whose members carry a non-Description attribute (with and without a Description).
+    private enum AttributedEnum
+    {
+        // Only a non-Description attribute -> should fall back to the member name.
+        [System.ComponentModel.Browsable(false)]
+        NonDescriptionOnly,
+
+        // A non-Description attribute alongside a Description -> the Description must still win,
+        // regardless of attribute ordering returned by reflection.
+        [System.ComponentModel.Browsable(false)]
+        [System.ComponentModel.Description("Mixed Description")]
+        MixedAttributes
+    }
+
     [TestInitialize]
     public void Setup()
     {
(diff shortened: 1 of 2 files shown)


tylerkron:
C:/Program Files/Git/agentic_review

tylerkron:
**BENCH RESULTS**

- Unit gate (`dotnet test --filter "TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests"`): **643 passed / 0 failed** (+3 new `EnumDescriptionConverter` tests).
- App-launch smoke (`Launch_MainWindowAppears_AndIsResponsive`): **passed**.
- Device bench: **skipped** — the change hardens display of undefined/malformed enum values in a value converter; it has no device-observable behavior reproducible on the bench (would require a device reporting an out-of-range WifiMode/WifiSecurityType).

tylerkron:
Ready for review: Qodo review is clean (Bugs 0, Rule violations 0), CI green (build-and-test + validate-pr-title pass), no unresolved review threads. Not merging — awaiting maintainer review.

PR #781, opened 2026-07-27
fix(connect): classify Core's typed SCPI-init error so it stops reaching Sentry

### LINKED ISSUE #775: bug: SCPI-init connect-classification arm is dead code against Core 1.3.0 (typed ScpiInitializationErrorException no longer derives from InvalidOperationException)
## Summary
The `LogConnectFailure` classification that downgrades a SCPI-init error to a Warning (issues #589/#709/#726/#733) is **dead code against the pinned `Daqifi.Core` 1.3.0**, so the environmental "device answered but rejected an init command" case still reports to Sentry as an app Error — the exact outcome those fixes intended to prevent.

## Root cause
Both `SerialStreamingDevice.LogConnectFailure` and `DaqifiStreamingDevice.LogConnectFailure` classify with:

```csharp
case InvalidOperationException when IsScpiInitializationError(ex):
```

where `IsScpiInitializationError` matches the message substrings `"SCPI error during initialization"` / `"SCPI error while setting stream interface to USB"`.

But in **Core 1.3.0** (which this repo pins — `<PackageReference Include="Daqifi.Core" Version="1.3.0" />`), those two throw sites use a **typed** exception introduced in Core #317:

- `Daqifi.Core/Device/DaqifiDevice.cs:1265` → `throw new ScpiInitializationErrorException("Device returned a SCPI error during initialization: ...")`
- `Daqifi.Core/Device/DaqifiStreamingDevice.cs:296` → `throw new ScpiInitializationErrorException("Device returned a SCPI error while setting stream interface to USB.")`

and crucially:

```csharp
// Daqifi.Core/Device/ScpiInitializationErrorException.cs:20
public class ScpiInitializationErrorException : Exception   // NOT InvalidOperationException
```

The C# type pattern `case InvalidOperationException when …` requires `ex is InvalidOperationException` **before** the `when` guard is evaluated. `ScpiInitializationErrorException` is not an `InvalidOperationException`, so the arm never matches, `IsScpiInitializationError` is never reached, and control falls through to `default:` → `AppLogger.Error(...)` → `SentrySdk.CaptureException`.

The exception propagates unwrapped (`InitializeAsync().GetAwaiter().GetResult()` → `catch (Exception ex) { LogConnectFailure(ex); }`), so nothing re-wraps it into an `InvalidOperationException`.

This is invisible to a green build (runtime type-pattern miss, not a compile error).

## Not affected (verified)
- `TransportNotConnectedException : InvalidOperationException` — the transport-disconnected/closed arms (#724/#741) still match correctly.
- `TcpStreamTransport` connect timeout is a genuine `System.TimeoutException` — the timeout arms are fine.

## Suggested fix
Match the typed exception directly (Core added it in 1.3.0 specifically so the condition can be classified without a message match):

```csharp
case ScpiInitializationErrorException:
    AppLogger.Warning(ex, "... returned a SCPI error during initialization ...");
    break;
```

`IsScpiInitializationError` then becomes unused and can be removed.

## Provenance
Found while porting this family to the Avalonia port (daqifi-avalonia), where an adversarial audit against the pinned Core 1.3.0 DLL flagged the dead arm. The port has taken the typed-exception fix; filing here since this repo carries the same dead arm on the same Core version.

tylerkron:
Closes #775.

**Not merging — opened for your review.**

## The bug

Both connect paths downgrade a SCPI-error-during-initialization to a Warning (so it does *not* capture to Sentry) with:

```csharp
case InvalidOperationException when IsScpiInitializationError(ex):
```

Against the pinned `Daqifi.Core` **1.3.0**, that arm is unreachable. Core throws its **typed** `ScpiInitializationErrorException` (daqifi-core#317) from both sites, and that type derives from `Exception`, **not** `InvalidOperationException`. A C# type pattern tests the type *before* the `when` guard runs, so the guard was never evaluated, control fell through to `default:`, and the environmental "device answered but rejected an init command" case kept hitting `AppLogger.Error` → `SentrySdk.CaptureException` — the exact outcome #589/#709/#726/#733 were meant to prevent.

Invisible to a green build: a runtime type-pattern miss, not a compile error.

## Verified against the pinned assembly, not assumed

I checked the actual `Daqifi.Core` 1.3.0 DLL rather than trusting the report:

| Check | Result |
| --- | --- |
| `ScpiInitializationErrorException` base chain | `Exception -> Object` — **not** an `InvalidOperationException` |
| `TransportNotConnectedException` base chain | `InvalidOperationException -> SystemException -> Exception` — so the transport arms (#724/#741/#588/#740) were **never** affected |
| Which ctor receives each SCPI message (IL scan) | `InitializeAsync`: `IL_0273 ldstr → IL_0283 newobj ScpiInitializationErrorException`<br>`OnDeviceInitializingAsync`: `IL_018D ldstr → IL_0195 newobj ScpiInitializationErrorException` |

The IL scan also confirms those are the **only** two methods in Core carrying a "SCPI error" string, so nothing else can still throw an `InvalidOperationException` with that wording — which is what makes deleting the message predicate safe rather than merely tidy.

## The fix

- Match `case ScpiInitializationErrorException:` in `SerialStreamingDevice.LogConnectFailure` and `DaqifiStreamingDevice.LogConnectFailure`.
- Delete the now-unreachable `AbstractStreamingDevice.IsScpiInitializationError` message predicate.

No behavior changes for any other arm.

## The tests were the real hole

The previous tests asserted the message predicate in isolation, plus `LogConnectFailure ... _DoesNotThrow`. Not throwing is true of *every* switch arm, so those tests stayed green the entire time the arm was dead — they could never have caught this.

So this PR also makes the classification observable:

- `AbstractStreamingDevice.AppLogger` becomes an **`init`-only `IAppLogger`** (was a get-only concrete `AppLogger`). A derived test double can inject a fake in its own constructor; production code still cannot swap it after construction. `Mock<IAppLogger>` is already the established pattern in this repo (`BootloaderWatcherTests`, `SessionDataRepositoryTests`), and `FirewallManager` already types its logger field as `IAppLogger`.
- Every `LogConnectFailure_*` case now asserts the **log level actually used**: `Warning` carrying the exception, and `Error` *never* called — plus the reverse guard that a genuine app bug (`InvalidOperationException("Transport exploded.")` / `"Connected Core device does not support streaming operations."`) still *does* reach `Error`, so "always warn" cannot pass either.
- The 7 message-predicate tests for the deleted predicate are removed; the `IsTransportClosedError` / `IsTransportDisconnectedError` predicate tests are untouched.

### Non-vacuity proved

Reverted the two arms to the pre-fix form and re-ran the fixtures: **3 failed / 11 passed**, failing exactly on the SCPI cases with

```
Expected invocation on the mock once, but was 0 times: l => l.Warning(ScpiInitializationErrorException: ..., It.IsAny<string>())
      IAppLogger.Error(ScpiInitializationErrorException: Device returned a SCPI error during initialization: -200,"Execution error", "Failed to connect on COM_TEST_589")
```

That `IAppLogger.Error(...)` line is the Sentry path, reproduced at runtime. Restoring the fix returns **14/14**.

## Gates

- `dotnet build` — 0 errors, **0 warnings**
- Unit gate (`TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests`) — **745 passed / 0 failed**
- App-launch smoke (`Launch_MainWindowAppears_AndIsResponsive`) — **PASS**, 15 s
- Device bench — **skipped, no board attached** (details in the BENCH RESULTS comment)

### DIFF
--- a/Daqifi.Desktop.Test/Device/SerialStreamingDeviceLogConnectFailureTests.cs
+++ b/Daqifi.Desktop.Test/Device/SerialStreamingDeviceLogConnectFailureTests.cs
@@ -17,6 +17,13 @@ namespace Daqifi.Desktop.Test.Device;
 /// asserted "did not throw", which is true of every switch arm, so they stayed green while the SCPI
 /// arm was unreachable (issue #775) and the condition was silently reaching Sentry.
 /// </para>
+/// <para>
+/// Each case verifies BOTH <see cref="IAppLogger.Error(Exception, string)"/> and
+/// <see cref="IAppLogger.Error(string)"/>, because both capture to Sentry — the exception overload
+/// via <c>SentrySdk.CaptureException(ex, ...)</c> and the message-only overload via a synthesized
+/// <c>AppLogErrorException</c>. Guarding only the exception overload would leave a live Sentry path
+/// a regression could take with these tests still green.
+/// </para>
 /// </summary>
 [TestClass]
 public class SerialStreamingDeviceLogConnectFailureTests
@@ -87,6 +94,7 @@ public class SerialStreamingDeviceLogConnectFailureTests
         // Error path (which is what captures to Sentry). Regression guard for #775.
         logger.Verify(l => l.Warning(ex, It.IsAny<string>()), Times.Once);
         logger.Verify(l => l.Error(It.IsAny<Exception>(), It.IsAny<string>()), Times.Never);
+        logger.Verify(l => l.Error(It.IsAny<string>()), Times.Never);
     }
 
     [TestMethod]
@@ -104,6 +112,7 @@ public class SerialStreamingDeviceLogConnectFailureTests
         // Assert
         logger.Verify(l => l.Warning(ex, It.IsAny<string>()), Times.Once);
         logger.Verify(l => l.Error(It.IsAny<Exception>(), It.IsAny<string>()), Times.Never);
+        logger.Verify(l => l.Error(It.IsAny<string>()), Times.Never);
     }
 
     [TestMethod]
@@ -122,6 +131,7 @@ public class SerialStreamingDeviceLogConnectFailureTests
         // Assert
         logger.Verify(l => l.Warning(ex, It.IsAny<string>()), Times.Once);
         logger.Verify(l => l.Error(It.IsAny<Exception>(), It.IsAny<string>()), Times.Never);
+        logger.Verify(l => l.Error(It.IsAny<string>()), Times.Never);
     }
 
     [TestMethod]
@@ -139,6 +149,10 @@ public class SerialStreamingDeviceLogConnectFailureTests
         // Assert
         logger.Verify(l => l.Error(ex, It.IsAny<string>()), Times.Once);
         logger.Verify(l => l.Warning(It.IsAny<Exception>(), It.IsAny<string>()), Times.Never);
+
+        // ...and via the exception-carrying overload specifically: the message-only Error(string)
+        // synthesizes an AppLogErrorException, which would strand the real stack trace out of Sentry.
+        logger.Verify(l => l.Error(It.IsAny<string>()), Times.Never);
     }
 
     /// <summary>
--- a/Daqifi.Desktop.Test/Device/SerialStreamingDeviceLogConnectFailureTests.cs
+++ b/Daqifi.Desktop.Test/Device/SerialStreamingDeviceLogConnectFailureTests.cs
@@ -1,82 +1,38 @@
+using Daqifi.Desktop.Common.Loggers;
 using Daqifi.Desktop.Device.SerialDevice;
+using Moq;
+using ScpiInitializationErrorException = Daqifi.Core.Device.ScpiInitializationErrorException;
 
 namespace Daqifi.Desktop.Test.Device;
 
 /// <summary>
-/// Tests for <see cref="SerialStreamingDevice"/>'s connect-failure classification. Certain
-/// InvalidOperationExceptions raised during connect are device/environmental conditions, not app
-/// bugs, so they must be downgraded to a Warning (no Sentry capture) rather than the default Error
-/// path: Core's SCPI-error-during-initialization message (issue #589) and the serial transport
-/// reporting the COM port closed mid-initialization (issue #588).
+/// Tests for <see cref="SerialStreamingDevice"/>'s connect-failure classification. Certain failures
+/// raised during connect are device/environmental conditions, not app bugs, so they must be
+/// downgraded to a Warning (which does NOT capture to Sentry) rather than the default Error path:
+/// Core's SCPI-error-during-initialization (issues #589/#775) and the serial transport reporting the
+/// COM port closed mid-initialization (issue #588).
+/// <para>
+/// The <c>LogConnectFailure_*</c> cases assert the log level actually used, via an injected
+/// <see cref="IAppLogger"/>. That is the whole point: the earlier versions of these tests only
+/// asserted "did not throw", which is true of every switch arm, so they stayed green while the SCPI
+/// arm was unreachable (issue #775) and the condition was silently reaching Sentry.
+/// </para>
 /// </summary>
 [TestClass]
 public class SerialStreamingDeviceLogConnectFailureTests
 {
-    // Exact wording Daqifi.Core 1.1.1 throws from DaqifiStreamingDevice.InitializeAsync()
-    // (verified against the NuGet package's DLL strings) when any command in its init sequence
-    // gets back a SCPI -200 execution error — e.g. "SYSTem:STReam:INTerface 0" (setting the
-    // stream interface to USB) rejected because firmware persisted WiFi as the last interface.
+    // Exact wording Daqifi.Core throws from DaqifiStreamingDevice.InitializeAsync() when any command
+    // in its init sequence gets back a SCPI -200 execution error — e.g. "SYSTem:STReam:INTerface 0"
+    // (setting the stream interface to USB) rejected because firmware persisted WiFi as the last
+    // interface.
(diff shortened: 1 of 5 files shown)


tylerkron:
## BENCH RESULTS — `15d062e`

| Gate | Result |
| --- | --- |
| `dotnet build` (solution) | **0 errors / 0 warnings** |
| Unit gate<br>`TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests` | **745 passed / 0 failed**, 1 m 34 s |
| App-launch smoke<br>`Launch_MainWindowAppears_AndIsResponsive` | **PASS**, 15 s |
| Device bench (RequiresDevice UITest) | **SKIPPED — no board attached** |

### Device-bench skip reason

Not a convenience skip and not a scope skip — the hardware is genuinely absent this run. Checked two ways before deciding:

- `[System.IO.Ports.SerialPort]::GetPortNames()` → **empty** (no COM ports at all)
- `Get-PnpDevice -PresentOnly` filtered on `VID_04D8` / `DAQiFi|Nyquist` / class `Ports` → **no matches**, so the board is not present in serial *or* HID-bootloader mode

Session was interactive (session 1) and **no DAQiFi process was running**, so nothing was competing for the port — the launch smoke ran clean on that same session.

### What was validated in place of the device

The changed classification only fires when a device answers and then rejects an init command, so a normal connect scenario would not have exercised it even with a board attached. It is instead pinned by runtime assertions:

- Reverting the two arms to the pre-fix form fails exactly the 3 SCPI cases, with the mock reporting `IAppLogger.Error(ScpiInitializationErrorException: ...)` — the Sentry path this PR removes, reproduced at runtime rather than argued statically. Restoring the fix returns 14/14.
- The one broadly device-facing part of this change is the `AppLogger` property becoming an `init`-only `IAppLogger` (it is on every device's logging path). That is covered by the launch smoke: the app boots, resolves the logger, and stays responsive.

Verification of the Core-side premise was done against the pinned `Daqifi.Core` 1.3.0 assembly (base-type reflection + an IL scan of the two throw sites), not against a running device — see the PR body table.

**Not merging — for your review.**

tylerkron:
﻿## BENCH RESULTS — `33681db` (test-only follow-up)

| Gate | Result |
| --- | --- |
| `dotnet build --no-incremental` (solution) | **0 errors / 0 warnings** |
| Unit gate<br>`TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests` | **745 passed / 0 failed**, 1 m 25 s |
| App-launch smoke<br>`Launch_MainWindowAppears_AndIsResponsive` | **PASS**, 12 s (interactive session, no DAQiFi process holding anything) |
| Device bench | **SKIPPED — two independent reasons, both stated rather than assumed** |

**Device-bench skip, part 1 (scope):** test-only commit. `git diff HEAD -- Daqifi.Desktop/` is empty, so production is byte-identical to `15d062e`.

**Device-bench skip, part 2 (availability):** re-probed this run — `[System.IO.Ports.SerialPort]::GetPortNames()` returned **zero ports**, and `Get-PnpDevice -PresentOnly` filtered on `VID_04D8` / `DAQiFi|Nyquist` returned **no matches**, so the board is absent in serial *and* HID-bootloader mode. Same as the previous run on this PR.

Test count unchanged at **745** — no test was added; five existing cases gained an assertion.

---

## Disposition of the Qodo review on `15d062e`

`Bugs (1) · Rule violations (0) · Skill insights (0)`, with **0 inline threads** — the single finding is summary-only, so this top-level comment is the reply.

### "Error overload untested" — **VALID, fixed in `33681db`**

Verified before acting rather than taken on faith:

- `IAppLogger` really does declare both `Error(string)` and `Error(Exception, string)` (`Daqifi.Desktop.Common/Loggers/IAppLogger.cs`).
- `AppLogger.Error(string)` really does reach Sentry — `SentrySdk.CaptureException(new AppLogErrorException(message))`, a dedicated type so message-only errors get their own Sentry grouping instead of colliding with every other generic `Exception`.

So the fixture's whole premise — "a Warning-classified connect failure must not reach Sentry" — was only half-guarded.

**The hole is narrower than it first looks, and I checked which regressions it actually catches.** A straight *swap* (`Warning(ex, ...)` → `Error(msg)`) was already caught by the existing `Warning(ex, ...) Times.Once`. The genuinely uncovered shape is an *additive* one: warn **and also** call the message-only Error. I reproduced exactly that — added an `AppLogger.Error(...)` next to the Warning in the SCPI arm and rebuilt:

- `Warning(ex, It.IsAny<string>()) Times.Once` — still passed
- `Error(It.IsAny<Exception>(), It.IsAny<string>()) Times.Never` — still passed
- the new `Error(It.IsAny<string>()) Times.Never` — **failed** on both SCPI cases: `Expected invocation on the mock should never have been performed, but was 1 times: l => l.Error(It.IsAny<string>())`

Restored, `grep -c` for the temp marker back to 0, clean rebuild 0/0. So the new assertion is load-bearing, not decorative.

**Also took the "optionally" half of the suggestion**, for a reason worth naming: on the two Error-classified cases, `Error(It.IsAny<string>()) Times.Never` pins that the *exception-carrying* overload is the one used — otherwise a regression to `Error(msg)` would strand the real stack trace outside Sentry behind a synthesized `AppLogErrorException`. That is a correctness claim about the Sentry payload, not just symmetry.

Coverage: 5 Warning-classified cases + 2 Error-classified cases across both fixtures, plus a `<para>` in each class doc recording *why* both overloads are verified, so the next person does not read the second line as redundant.

**Relevance line checked, not skimmed** (fire-log lesson: when Qodo says the team accepted a class of change, verify before dismissing). It cited #733/#447 as prior "strengthen the connect-failure Sentry guard" work — consistent with this PR's own stated purpose, so there was nothing to argue.

---

**Not merging — for review.**

Issue #708, opened 2026-07-16
tracking: desktop deletions blocked on Daqifi.Core work (~430 lines across 8 Core tickets)


tylerkron:
## Purpose

Umbrella for desktop code that is deletable **once a corresponding Core change lands**. Each row is a Core ticket carrying the full rationale and the desktop file/line refs; this issue exists so the desktop-side deletions aren't invisible from this repo. Close each row here as its Core ticket ships and we take the version bump.

Filed as one tracking issue rather than eight stubs, since none of these have desktop-side work until Core moves.

| Core ticket | Desktop code deleted | ~Lines |
|---|---|---|
| daqifi-core#302 - friendly-name SCPI + validation + `DeviceMetadata.FriendlyName` **(shipped - merged daqifi-core#313, not yet released)** | `AbstractStreamingDevice.cs:80-84`, `:606-641`, `:1768-1828` | 85 |
| daqifi-core#309 - keep `IChannel` instances stable across re-population **(shipped - merged daqifi-core#314, not yet released; confirmed Core was recreating instances every status message, now fixed)** | `AnalogChannel.cs:153-173`, `DigitalChannel.cs:201-230`, `AbstractStreamingDevice.cs:124-125`, `:781-801`, `:1549-1613`, `:1620-1630` | 110 |
| daqifi-core#304 - `WifiBridgeActivator.Deactivate` | `FirmwareUpdateCoordinator.cs:476-535` | 50-60 |
| daqifi-core#303 - `SdCardLogEntry.HasDeviceTimestamp` | `Loggers/ImportTimestampQuality.cs` + call sites at `SdCardSessionImporter.cs:211`, `:284-290` | 60 |
| daqifi-core#308 - classified `StatusMessageReceived` / `StreamMessageReceived` events | `AbstractStreamingDevice.cs:114`, `:502-505`, `:579-588`, `:594-604`, `:901-904`, `:1714-1718` | 45 |
| daqifi-core#306 - commandable PWM freq/duty defaults + clamping | `AbstractStreamingDevice.cs:86-93`, `:1428-1446`; `DigitalChannel.cs:9-17`, `:132-153`, `:189-199` | 35 |
| daqifi-core#305 - `DeviceMetadata.CopyFrom` / `DeviceCapabilities.Clone` | `AbstractStreamingDevice.cs:1652-1668`, `:1697-1710` | 28 |
[...]

Issue #707, opened 2026-07-16
chore: delete desktop tests that exercise Core's DeviceTypeDetector and bare arithmetic


tylerkron:
## Change

Delete ~110 lines of tests in `Daqifi.Desktop.Test/Device/AbstractStreamingDeviceTests.cs` that exercise Core (or bare arithmetic) rather than any desktop code:

- `:21-90` - `RemoveChannelBitOperation_ShouldUseLeftShiftNotRightShift`, `ChannelSetByte_IntegerOverflow...`, `StringConversion_ChannelBitMask...`
- `:345-425` - `DeviceType_Enum_ShouldHaveCorrectValues`, `DeviceType_EnumNames...`, and all seven `DetectFromPartNumber_*` tests

The `DetectFromPartNumber_*` and `DeviceType_*` tests cover Core's `DeviceTypeDetector` / `DeviceType`. The bit-operation tests assert on plain shift/overflow arithmetic with no desktop call in them.

## Why bother

These are worse than useless: they cost coverage-gate time, and they give false confidence that *we* test part-number detection when in fact a Core regression is what they'd catch - in the wrong repo, one release late.

If any of them cover a Core behavior daqifi-core doesn't already test, port it there before deleting rather than dropping the coverage on the floor.

_From a cross-repo audit to make daqifi-desktop a thin wrapper over Daqifi.Core (2026-07-16, Core 1.1.1)._

Issue #613, opened 2026-06-19
refactor: Delegate stream-frame decoding to Core's per-channel sample pipeline


tylerkron:
## Change
Remove the hand-rolled stream-frame demux in `ProcessStreamMessage` (`Daqifi.Desktop/Device/AbstractStreamingDevice.cs:501-639`): active-channel ordering, the USB-float-vs-WiFi-raw branch, and digital bit unpacking. Subscribe to Core's per-channel sample events instead, and keep only a thin map from Core's sample to the desktop's richer `DataSample` (DeviceName/SerialNo/Color/FirmwareDeltaMs/LoggingSessionID).

## Dependency
**Blocked by daqifi-core#242** (decoded per-frame sample pipeline). The desktop keeps its `DataSample` mapping layer; Core must surface enough per-sample metadata for a cheap map.

## Related
- daqifi-desktop#449 (parse `analog_in_data_ts` for batched PB messages) — should be absorbed by the Core pipeline.

_From a cross-repo audit to make daqifi-desktop a thin wrapper over Daqifi.Core._

tylerkron:
**Unblocked.** daqifi-core#242 (decoded per-frame sample pipeline) closed 2026-07-03 and shipped in Core 1.0.0 (consumed since #669). Core's `DaqifiStreamingDevice` now owns `DecodeStreamFrame` (analog float-vs-raw branch, digital bit unpacking, active-channel ordering) and raises per-channel `SampleReceived` events, so the hand-rolled demux in `ProcessStreamMessage` (now at `AbstractStreamingDevice.cs:535-663`) can shrink to the thin map into the desktop's richer `DataSample` as this issue describes.

PR #786, opened 2026-07-27
chore(device): subscribe Core's classified message events

tylerkron:
Row 5 of the [#708](https://github.com/daqifi/daqifi-desktop/issues/708) deletion backlog. daqifi-core#308 shipped in Core **1.2.0**; we're on **1.3.0**.

Core already classifies every inbound protobuf frame — that's how it decides whether to update metadata and populate channels. The desktop was subscribing to the undifferentiated `MessageReceived` and running a **second** `ProtobufProtocolHandler` pass over the same frame to re-derive the answer Core had just computed. At streaming rates, on every frame.

**Deleted:** the `_protocolHandler` field, `InitializeProtocolHandler`, `OnInboundMessageReceived`, `OnCoreMessageReceived`, `HandleInboundMessage`. `InitializeDeviceState` existed only to build that handler, so it goes from `AbstractStreamingDevice` and from `IStreamingDevice` (sole implementer). Subscription is now a `SubscribeCoreDeviceEvents` mirroring the `UnsubscribeCoreDeviceEvents` that already existed.

## This also fixes a latent correctness hazard

`_acceptChannelSamples` gates which decoded samples reach desktop channels. It's set at the end of `ProcessStreamMessage` and must still be set when Core decodes *that same frame*.

- **Before:** the desktop was reached via `MessageReceived` → `_ = _protocolHandler.HandleAsync(...)` — **fire-and-forget**. It worked only because that task happened to complete inline. An `await` introduced anywhere inside that path would have silently decoupled the gate from the frame it guards, and the symptom would have been wrong samples on the wrong frame, not a crash.
- **After:** Core raises the classified event *synchronously* from inside `DaqifiStreamingDevice.OnStreamMessageReceived`, immediately before its own `DecodeStreamFrame`. The ordering is now structural.

As a bonus, Core wraps classified-event subscribers in `RaiseClassifiedEvent`, which catches and logs a throwing subscriber. Previously an exception on this path vanished into the discarded task.

`OnStatusMessageReceived`/`OnStreamMessageReceived` become `protected` — they replace `HandleInboundMessage` as the test seam and name what they actually are: the handlers Core invokes for an already-classified frame.

## Testing

`dotnet test --filter "TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests"` → **768/768 pass**.

Test fixtures that previously drove both halves by hand (`HandleInboundMessage` *then* `SimulateStreamFrame`) now make one call to Core's own `OnStreamMessageReceived`, which raises the classified event and then decodes. That's closer to production than the two-step simulation it replaces — the ordering is Core's rather than something each fixture has to reproduce correctly.

Tests named for the classification (`RouteInboundMessage_StatusMessage…`) now call the specific handler, since which handler a frame reaches is Core's decision, not ours — consistent with #707's "don't test Core here". The recorded knowledge about which frame shapes classify which way is preserved in the comments rather than dropped.

One thing worth a reviewer's eye: while converting the fixtures I hit a real double-dispatch — a fixture that both invoked the desktop handler directly *and* was subscribed to Core processed every frame twice. Caught by `FrameTimestampSourceTests`. Worth confirming I didn't leave a similar one anywhere.

⚠️ **Streaming hot path** — the unit suite covers the gating logic well, but a smoke test against real hardware (stream, check samples land on the right channels with sane timestamps) would be worth it before merge.

Independent of #783/#784/#785.

Refs #708

### DIFF
--- a/Daqifi.Desktop.Test/Device/AbstractStreamingDeviceTests.cs
+++ b/Daqifi.Desktop.Test/Device/AbstractStreamingDeviceTests.cs
@@ -26,20 +26,17 @@ public class AbstractStreamingDeviceTests
     }
 
     [TestMethod]
-    public void RouteInboundMessage_StatusMessageWithFriendlyDeviceName_UpdatesFriendlyName()
+    public void StatusMessageWithFriendlyDeviceName_UpdatesFriendlyName()
     {
         // Arrange
         var device = new TestStreamingDevice();
-        device.InitializeDeviceState(); // wires the protocol handler that classifies/routes inbound messages
 
         // Act — firmware's fast streaming-frame encoder (Nanopb_EncodeStreamingFast) hardcodes
         // only msg_time_stamp/analog_in_data/digital_data/digital_port_dir; friendly_device_name
         // is only ever populated on "info" responses like SYSTem:SYSInfoPB? (Core sends this once
-        // during InitializeAsync). Core's ProtobufProtocolHandler classifies a message with no
-        // analog/digital sample data but a nonzero port-count field as "Status", routed to
-        // OnStatusMessageReceived — not OnStreamMessageReceived (see ProtobufProtocolHandler.
-        // IsStatusMessage/IsStreamMessage).
-        device.RouteInboundMessage(new DaqifiOutMessage
+        // during InitializeAsync). Core classifies a message with no analog/digital sample data but
+        // a nonzero port-count field as Status and raises StatusMessageReceived for it.
+        device.RouteStatusMessage(new DaqifiOutMessage
         {
             AnalogInPortNum = 1,
             DeviceSn = 12345,
@@ -52,16 +49,15 @@ public class AbstractStreamingDeviceTests
     }
 
     [TestMethod]
-    public void RouteInboundMessage_StreamMessageWithFriendlyDeviceName_UpdatesFriendlyName()
+    public void StreamMessageWithFriendlyDeviceName_UpdatesFriendlyName()
     {
         // Arrange — belt-and-suspenders: a real Stream-classified frame never carries this field
         // (see the Status-message test above), but the desktop code captures it there too in
         // case firmware's streaming field set ever changes.
         var device = new TestStreamingDevice();
-        device.InitializeDeviceState();
 
         // Act
-        device.RouteInboundMessage(new DaqifiOutMessage
+        device.RouteStreamMessage(new DaqifiOutMessage
         {
             MsgTimeStamp = 1000,
             DeviceSn = 12345,
@@ -74,15 +70,17 @@ public class AbstractStreamingDeviceTests
     }
 
     [TestMethod]
-    public void RouteInboundMessage_StatusMessageWithoutFriendlyDeviceName_ClearsStaleFriendlyName()
+    public void StatusMessageWithoutFriendlyDeviceName_ClearsStaleFriendlyName()
     {
         // Arrange — e.g. SerialStreamingDevice instances are reused across reconnects on the same
         // COM port (ConnectionDialogViewModel dedups discovery by port), so a name captured from a
         // previously connected device must not leak onto a different/renamed device that reports
         // no name (issue #83 Qodo review: "stale friendlyname leaks").
+        //
+        // This is also why FriendlyName is not folded into Core's Metadata.FriendlyName: Core's
+        // UpdateFromProtobuf assigns the name only when non-empty, so it can never clear.
         var device = new TestStreamingDevice();
-        device.InitializeDeviceState();
-        device.RouteInboundMessage(new DaqifiOutMessage
+        device.RouteStatusMessage(new DaqifiOutMessage
         {
             AnalogInPortNum = 1,
             DeviceSn = 12345,
@@ -92,7 +90,7 @@ public class AbstractStreamingDeviceTests
 
         // Act — a fresh connect's status response with no name (firmware always includes the
         // field, empty when unset) is authoritative and must clear the stale value.
-        device.RouteInboundMessage(new DaqifiOutMessage
+        device.RouteStatusMessage(new DaqifiOutMessage
         {
             AnalogInPortNum = 1,
             DeviceSn = 67890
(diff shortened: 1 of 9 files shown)


tylerkron:
### Response to Qodo review

**"Test methods contain underscores" (rule 244817) — declined, as previously.**

`AbstractStreamingDeviceTests.cs` follows the standard MSTest
`MethodUnderTest_Scenario_ExpectedResult` naming convention, which is used consistently across
all ~770 tests in `Daqifi.Desktop.Test`. The underscores are the convention's segment separators,
not word separators inside an identifier, and they are what makes a failing test name readable in
CI output (`StatusMessageWithoutFriendlyDeviceName_ClearsStaleFriendlyName` says what broke;
`StatusMessageWithoutFriendlyDeviceNameClearsStaleFriendlyName` does not).

This is the same finding declined on #739, #727 and #748 — Qodo itself rates it Low relevance and
cites those PRs. Renaming these tests would make this file inconsistent with every other test file
in the project. No change made.

### Second-opinion review of the diff

Qodo found 0 bugs here, so the diff got an independent re-read focused on the failure modes this
kind of change actually has. Verified against the Core source at the consumed tag (`v1.3.0`):

- **No double-subscription, no leaked subscription.** `SubscribeCoreDeviceEvents` has exactly one
  production call site (`Connect()`), which is preceded by `CleanupConnection()`; every `+=` has a
  matching `-=` on the same event source in `UnsubscribeCoreDeviceEvents`, reached from both
  `Disconnect()` and `CleanupConnection()`. Method-group `-=` matches by target+method, so the
  fresh delegate instance is not a problem. Neither transport overrides `Connect()`, and
  `SyncFromCoreDevice` does not also subscribe `ChannelsPopulated`.
- **No double-processing.** Core raises `MessageReceived` from exactly two places
  (`DaqifiDevice.OnStatusMessageReceived` line 1438 and `OnStreamMessageReceived` line 1746), each
  immediately after the classified event for the same frame — so the old raw path and the new
  classified path were always the same two message classes, and the old path re-derived the
  classification with the identical `ProtobufProtocolHandler.DetectMessageType`. Dropping it is
  exact routing parity, not a second delivery. `MessageReceived` now has no desktop subscriber.
- **Ordering and threading are unchanged.** The handler still runs on Core's consumer thread, one
  link earlier in the *same synchronous* call chain (`RaiseClassifiedEvent` → handler → 
  `OnMessageReceived`), still inside `base.OnStreamMessageReceived` and therefore still strictly
  before `DecodeStreamFrame`. No new thread affinity, so nothing new needs marshalling. For status
  frames, `FriendlyName` is still assigned after Core's `Metadata.UpdateFromProtobuf` /
  `PopulateChannelsFromStatus`, so `OnCoreChannelsPopulated` ordering is preserved.
- **Nothing was silently dropped.** The deleted code held no error classification or Sentry
  suppression — only an `AppLogger.Information` line and a no-op SD-card handler for a
  `ProtobufMessageType.SdCard` value `DetectMessageType` never returns. The Warning-level
  classification for transport/SCPI-init failures lives in `LogConnectFailure`, untouched.

One behavioural nuance worth recording rather than changing: Core's `RaiseClassifiedEvent` catches
a subscriber exception, where the old route let it propagate into `StreamMessageConsumer`'s
per-message `catch` → `OnErrorOccurred`. Both swallow, and the new one is finer-grained (the frame's
samples are still dropped, because `_acceptChannelSamples` stays false). Not a regression.

### Changes made (7f8da9d)

The subscribe/unsubscribe symmetry was correct but **untested** — the one way this change could
leak. `CoreConnectionTemplateTests` already asserted it for `ChannelsPopulated`; it now covers both
classified events too (`Disconnect_UnsubscribesCoresClassifiedMessageEvents`). Confirmed
load-bearing by mutation: deleting either `-=` fails the matching assertion.

Also fixed two documentation artifacts of the change itself:
- `FrameTimestampSourceTests.cs` — a stale duplicate `<summary>` left on
  `FrameTimestampTestDevice.RouteStreamFrame`, still describing the removed two-step routing.
- `docs/architecture.md` — the streaming sequence diagram still showed
  `HandleInboundMessage → ProtobufProtocolHandler.HandleAsync`, a step this PR deletes.

Build and the full unit gate are green (769/769).

tylerkron:
### Response to Qodo review (round 2) — 38af8d7

**1. `message` parameter docs missing (rule 244813) — fixed.**
`SimulateStatusMessage` / `SimulateStreamMessage` referenced `<paramref name="message"/>` in their
summaries but had no `<param>` element. Both now have one.

**2. Core decode missing from diagram (Bug) — fixed, and a good catch.**
This one was correct and I'd wrongly scoped it out. When I updated the sequence diagram I fixed the
step this PR invalidated (`HandleInboundMessage → ProtobufProtocolHandler.HandleAsync`) but left the
*next* step — which has been wrong since issue #613 moved channel decoding into Core, and which is
precisely the ordering contract this PR's rationale rests on. Having touched the diagram, leaving the
adjacent step attributing Core's decode to `AbstractStreamingDevice` was not defensible.

The diagram now shows both hops as production actually runs them:

```
Core->>Dev: StreamMessageReceived event (classified)
Dev->>Dev: OnStreamMessageReceived → ProcessStreamMessage    // timestamp + open the gate
Core->>Core: DecodeStreamFrame (same frame, once the event returns)
Core->>Dev: IChannel.SampleReceived (per channel)
Dev->>Ch:   channel.ActiveSample = new DataSample(...)
```

The WiFi-vs-USB scaling note now names `DaqifiStreamingDevice.DecodeStreamFrame` as the owner of the
`AnalogInData` vs `AnalogInDataFloat` branch instead of `OnStreamMessageReceived`, and a new note
states the gate contract the diagram now makes visible (Core decodes leftover frames too; the
`_acceptChannelSamples` gate is what drops them).

**3. Test methods contain underscores (rule 244817, Low) — declined again.**
Unchanged position from my previous comment: `MethodUnderTest_Scenario_ExpectedResult` is the MSTest
convention used consistently across all 769 tests in this project, and Qodo itself rates this Low
relevance citing #739 / #727 / #748 where it was declined. Renaming would make this one file
inconsistent with every other test file.

Build and the full unit gate are green (769/769).

Issue #805, opened 2026-08-03
reliability: subscribe to Core's ErrorOccurred and SendFailed so background failures stop vanishing


tylerkron:
## Summary

Core v1.4.0 made its background failures observable (daqifi-core#377 / #378 / #394 / #413 / #415). Two new event surfaces exist and **we subscribe to neither**, so these failures still vanish silently in the desktop app.

**Blocked by:** #802 (the Core 1.3.0 -> 1.4.0 bump).

## `IDevice.ErrorOccurred`

New in v1.4.0: `event EventHandler<DeviceErrorEventArgs> ErrorOccurred` on `IDevice` itself (`IDevice.cs:55`), with `DeviceErrorSource` classifying where the failure came from and a built-in `DeviceErrorThrottle` so a repeating background fault does not flood the sink.

Before this, a failure on Core's read loop or consumer threads had nowhere to go. The release notes describe the change as "background failures are visible, and the last silent read-loop spin is gone."

Wire it in `ConnectionManager` at the same place we already attach `ConnectionLost` (`Daqifi.Desktop/ConnectionManager.cs:285`), and unsubscribe at the two teardown sites (`:311`, `:348`) - the friendly-name handler leak fixed in #795 is the pattern to avoid repeating.

**Routing matters more than wiring here.** We have a repeated history of environmental conditions landing on `AppLogger.Error` and becoming Sentry noise (#775, #779, #801). Classify by `DeviceErrorSource` and default to `AppLogger.Warning`; reserve `Error` for sources that genuinely indicate an app bug. Lean on Core's throttle rather than adding our own.

## `IMessageProducer.SendFailed`

New in v1.4.0: `event EventHandler<MessageSendFailedEventArgs<string>> SendFailed` (daqifi-core#413). Producer write failures previously vanished - a SCPI command could fail to reach the device with no error, no log, and no user-visible signal, leaving the app's idea of device state silently diverging from the device's.

`DaqifiDevice` re-exposes it at `DaqifiDevice.cs:521`. Subscribe alongside `ErrorOccurred`.

Worth checking whether any existing "device ignored my command" symptom is explained by this. #589 (SCPI error while setting stream interface to USB) is a plausible candidate to re-examine once we can see send failures.

## Also available, lower priority

[...]

Issue #804, opened 2026-08-03
feat: adopt Core's auto-reconnect and streaming session resume


tylerkron:
## Summary

Core v1.4.0 ships opt-in auto-reconnect with streaming session resume (daqifi-core#379 / #418). A device that drops mid-capture can reconnect and pick the stream back up. We do not use it.

Today a spontaneous transport drop is terminal: `ConnectionManager.OnDeviceConnectionLost` (`Daqifi.Desktop/ConnectionManager.cs:381`) unsubscribes the device's channels, tears the connection down via `Disconnect`, and raises a notification. The user rebuilds the session by hand and the capture in progress is lost.

**Blocked by:** #802 (the Core 1.3.0 -> 1.4.0 bump).

## What Core provides

- `DaqifiDevice.ReconnectOptions` (`ConnectionManager.cs` would set this at connect time). Presets: `Disabled` (the default, so current behavior is preserved until we opt in), `Default`, `Fast`, `Resilient`. Tunables: `Enabled`, `MaxAttempts`, `InitialDelay`, `MaxDelay`, `BackoffMultiplier`, `ResumeStreaming`, plus `CalculateDelay(attemptNumber)`.
- Events: `ReconnectAttempt` (`ReconnectAttemptEventArgs`), `Reconnected` (`ReconnectedEventArgs`), `ReconnectFailed` (`ReconnectFailedEventArgs`).
- `IsReconnecting` and `CancelReconnect()`.
- `DeviceReconnectFailedException` for the terminal case.
- `ConnectionStatus.Retrying` for the intermediate state, which the existing `DAQiFiConnectionStatus` mapping at `ConnectionManager.cs:363-367` has no equivalent for.

Reconnect is also now genuinely reachable over USB: v1.4.0's `SerialStreamTransport` detects a physical unplug (write-probe plus port-presence poll, bounded at roughly three seconds) instead of trusting `SerialPort.IsOpen`, which stays open after the device is pulled.

## Design questions to settle before implementing

1. **Opt-in or on by default?** `ReconnectOptions.Disabled` is Core's default, so doing nothing keeps today's behavior. Suggest starting behind a setting, defaulting to on for `ResumeStreaming` during an active logging session only.
2. **What the UI shows while `IsReconnecting`.** The device tile currently flips to disconnected. A "reconnecting, attempt N of M" state needs a `DAQiFiConnectionStatus` member and a tile treatment. Follow docs/style-guide.md.
[...]

Issue #803, opened 2026-08-03
chore: retire the four daqifi-core#398 bandaids now that Core v1.4.0 owns them


tylerkron:
## Summary

daqifi-core#398 ("api: retire four downstream workarounds") is **closed and shipped in Core v1.4.0**. All four gaps it tracked are fixed upstream, and all four of the desktop bandaids that worked around them shipped here in the last two weeks (#776, #779, #780, #782). This issue deletes them.

This is the single largest thin-wrapper win available from the v1.4.0 bump: every row below is a deletion, not a rewrite.

**Blocked by:** #802 (the Core 1.3.0 -> 1.4.0 bump).

### 1. `TimestampProcessor.ResetAll` no longer drops device clock config (gap 3)

The bandaid: the `_timestampFrequencyApplied` gate plus the `_timestampProcessorSync` lock that exists to keep the apply and its gate write atomic against a UI-thread `ResetAll`.

- `Daqifi.Desktop/Device/AbstractStreamingDevice.cs:754` (the apply + reconstruction under lock)
- `:1222` (`InitializeStreaming` reset)
- `:1259` (`StopStreaming` reset)
- the `_timestampFrequencyApplied` field and the ~10 lines of comment at `:740-751` explaining why the lock is load-bearing
- `Daqifi.Desktop.Test/Device/TimestampProcessorSerializationTests.cs` in full, including the `RecordingTimestampProcessor` fake that #802 otherwise has to extend

Core v1.4.0 `TimestampProcessor.ResetAll()` clears session baselines only and preserves `_deviceTickPeriods`. Core's own source comment names our workaround as the thing it is retiring:

> Reset(deviceId) has always preserved them; ResetAll now matches (#398 gap 3). ... forced every consumer to re-apply the frequency after each reset behind its own "have I applied it yet" gate.

Once the frequency survives a reset, the apply happens once and there is no apply/gate/reset interleaving to serialize. Confirm before deleting the lock that nothing else has crept under `_timestampProcessorSync`.

[...]

Issue #779, opened 2026-07-27
bug: SD-download stall watchdog is dead code over USB serial — Core's plain TimeoutException hits the classifier's default (Sentry) arm


cptkoolbeenz:
## Summary
The #755 stall-degradation machinery does not trigger on the only transport SD import supports (USB serial). `SdCardFailureClassifier` matches `SdCardDownloadStalledException` **by type** (deliberately, to avoid misclassifying unrelated timeouts), and `SdCardSessionImporter.DownloadWithStallWatchdogAsync` only produces that exception from an `OperationCanceledException` raised by its own 90s watchdog. But over USB serial:

- Core's serial transport sets `SerialPort.ReadTimeout ≈ 500ms` and hands the raw `BaseStream` to `SdCardFileReceiver`.
- `.NET SerialStream.ReadAsync` returns **0 bytes** on a read timeout (it does not throw, and ignores the `CancellationToken` mid-read).
- `SdCardFileReceiver` treats `bytesRead == 0` as fatal and throws a **plain `TimeoutException`** ("Transport stream closed before receiving the EOF marker…").

So a device that wedges mid-transfer surfaces a plain `TimeoutException` in ~500ms — long before the 90s watchdog, and never as a cancellation. `DownloadWithStallWatchdogAsync`'s OCE-only catch misses it, `SdCardDownloadStalledException` is never produced, and the classifier's `default` arm fires: `LogError` (Sentry) + "check the connection" guidance + `IsCardUnavailable = false` so `ImportAllFiles` grinds through every remaining file. That's precisely the #754 regression this feature was written to eliminate, reproduced on the supported transport.

## Fix direction
Normalize the transport-shaped `TimeoutException` from the download call to the stalled condition. Catching `TimeoutException` **scoped to the `DownloadSdCardFileAsync` call** (where any timeout is by definition a stall) and re-throwing `SdCardDownloadStalledException` is robust (type-scoped, not message-matched) and leaves unrelated timeouts elsewhere on the Error path.

## Provenance
Found by an adversarial audit while porting #755 to the Avalonia port (daqifi-avalonia), where the port applied the scoped-`TimeoutException` normalization as a corrective fix. Filing here since the upstream classifier/watchdog share the gap.

PR #806, opened 2026-08-03
deps: bump Daqifi.Core to 1.4.0 and classify SerialPortConnectException

### LINKED ISSUE #801: SerialStreamingDevice.LogConnectFailure will route routine connect failures to Sentry once Core's SerialPortConnectException lands
## Summary

`SerialStreamingDevice.LogConnectFailure` classifies serial connect failures by exception type. daqifi-core PR #427 (closes daqifi-core#424) replaces the two types it matches on with a new `SerialPortConnectException`, so **both arms become dead code and routine connect failures fall through to the `default:` arm — `AppLogger.Error`, i.e. Sentry.**

This is the same failure mode as #775 and #779: Core changed a thrown type, a desktop classification arm silently stopped matching, and environmental conditions started being reported as app bugs.

## The code today

`Daqifi.Desktop/Device/SerialDevice/SerialStreamingDevice.cs:100`

```csharp
case FileNotFoundException:          // port no longer exists  -> Warning
case UnauthorizedAccessException:    // port held by another process -> Warning
...
default:
    AppLogger.Error(ex, $"Failed to connect on {PortName}");   // <-- Sentry
```

## What changes in Core

`SerialStreamTransport.ConnectAsync` now translates `SerialPort.Open()` failures into `SerialPortConnectException` (derives from `IOException`), carrying a typed `SerialPortConnectFailure` reason: `NotFound`, `InUse`, `AccessDenied`, `Unknown`. The original platform exception is always preserved as `InnerException`.

Neither `case FileNotFoundException:` nor `case UnauthorizedAccessException:` matches it. A missing or busy port — the two most common things a user hits — becomes a Sentry error.

## Why the fix is a strict improvement, not just a patch

The old branching could never have worked reliably anyway. Core's investigation found that on macOS a *missing* port and a *busy* port throw byte-identical exceptions, the inner `IOException` message was not stable across processes for the same port (`Unknown error: 203` vs `No such file or directory`), and the inner `HResult` is a stale errno. So desktop's two arms were matching on platform-specific accidents. `SerialPortConnectFailure` is a stable, cross-platform discriminator — this is the first time desktop can classify these correctly.

## Suggested fix

Add an arm before the existing ones:

```csharp
case SerialPortConnectException { Reason: SerialPortConnectFailure.NotFound }:
    AppLogger.Warning(ex, $"Cannot connect on {PortName}: port is not available");
    break;
case SerialPortConnectException { Reason: SerialPortConnectFailure.InUse }:
    AppLogger.Warning(ex, $"Cannot connect on {PortName}: port is in use by another process");
    break;
case SerialPortConnectException { Reason: SerialPortConnectFailure.AccessDenied }:
    AppLogger.Warning(ex, $"Cannot connect on {PortName}: access denied");
    break;
```

Keep the existing `FileNotFoundException`/`UnauthorizedAccessException` arms so the classification still works against older Core versions.

Worth extending `Daqifi.Desktop.Test/Device/SerialStreamingDeviceLogConnectFailureTests.cs` with a case per reason — that suite exists precisely to pin this classification.

## Related open Sentry issues this should quieten

- #796 — `System.UnauthorizedAccessException: Access is denied. (0x80070005)`. Per Core's finding this message is actively misleading: it is usually a *missing or stale* port, not a permissions problem.
- #798 — `System.IO.IOException: A device which does not exist was specified.`

## Timing

Not urgent until daqifi-core#427 merges and desktop takes the Core update — but it must land in the same release that picks up that Core version, or Sentry gets noisy.

_Found while implementing daqifi-core#424._

### LINKED ISSUE #802: chore: bump Daqifi.Core 1.3.0 -> 1.4.0
## Summary

[Daqifi.Core v1.4.0](https://github.com/daqifi/daqifi-core/releases/tag/v1.4.0) shipped 2026-08-03. Bump `Daqifi.Desktop/Daqifi.Desktop.csproj:41` from `1.3.0` to `1.4.0`.

Filed on its own because a batch of deletions and adoptions are blocked behind it and it should merge first.

## Measured blast radius

Trial-bumped and built (`dotnet build -tl:on`, 2026-08-03). Exactly **two** compile errors, both mechanical:

1. `Daqifi.Desktop/Device/Firmware/BootloaderSessionStreamingDeviceAdapter.cs:37` - CS0535, does not implement `IDevice.ErrorOccurred`. Add no-op `add`/`remove` accessors next to the existing `MessageReceived` ones, and extend the "intentionally no-op" comment block at `:81-86` to name the 1.4.0 addition the way that comment already tracks 0.24.0/1.0.0/1.3.0.
2. `Daqifi.Desktop.Test/Device/TimestampProcessorSerializationTests.cs:394` - CS0535, the `RecordingTimestampProcessor` fake does not implement the new `ITimestampProcessor.HasTimestampFrequency(string)`.

Everything else compiles clean, including through Core splitting `DaqifiStreamingDevice` and `FirmwareUpdateService` into collaborators (~2,400 lines moved out of `FirmwareUpdateService` alone) and extracting `SdCardOperations`. Unlike the 1.1.1 -> 1.3.0 bump, no `System.Management` pin floor change was needed.

## Notable API surface arriving with this bump

Not all of it needs adoption here - listed so the reviewer knows what changed underfoot:

- `ConnectionStatus` gained `Lost`, `Retrying`, `Failed`. `SerialStreamTransport` now actually detects a physical unplug (write-probe plus port-presence poll, bounded at roughly three seconds), so `Lost` is reachable over USB for the first time.
- `IDevice.ConnectAsync` / `DisconnectAsync` take a `CancellationToken`; `DaqifiDevice` implements `IAsyncDisposable`.
- New `IDevice.ErrorOccurred` (`DeviceErrorEventArgs` / `DeviceErrorSource` / `DeviceErrorThrottle`) and `IMessageProducer.SendFailed`.
- `ReconnectOptions` plus `ReconnectAttempt` / `Reconnected` / `ReconnectFailed` events and `CancelReconnect()`.
- `DeviceConnectionOptions` with `PreserveActiveStream` (attach without disturbing a stream already running).
- Per-device operation serialization via `RunExclusiveAsync`, so concurrent callers stop reading each other's replies.
- `DeviceNotConnectedException` replaces untyped connectivity guard failures. Cross-check against the pattern in #619 / #622 before assuming call sites are unaffected.
- `FirmwareUpdateState.ReconnectingAfterFlash = 12` is a **new enum member**; any exhaustive switch over `FirmwareUpdateState` needs an arm.

## Verification

- `dotnet build -tl:on`
- `dotnet test --filter "TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests" -tl:on`
- Bench smoke: connect over USB, stream, stop, SD list. The transport drop-detection change is the riskiest behavioral delta and is not covered by unit tests here.

tylerkron:
Bumps `Daqifi.Core` 1.3.0 -> 1.4.0 and takes the one change that **must** ship in the same PR as the bump.

Closes #802
Closes #801

## Why these are one PR

Core 1.4.0 (daqifi-core#427) translates `SerialPort.Open()` failures into a typed `SerialPortConnectException` carrying a stable `SerialPortConnectFailure` reason. It derives from `IOException`, so it matches **neither** the `FileNotFoundException` nor the `UnauthorizedAccessException` arm in `SerialStreamingDevice.LogConnectFailure`.

Bumping without the new arms means a missing port and a busy port - the two most common things a user hits - fall through to the `default:` arm, `AppLogger.Error`, i.e. Sentry. That is the regression #801 was filed to pre-empt.

## What changed

**The bump** (2 mechanical fixes, both predicted in #802):
- `BootloaderSessionStreamingDeviceAdapter` implements Core's `IDevice` directly, so it gains `ErrorOccurred`. No-op accessors, matching the existing `MessageReceived` treatment. Its version-history comment is extended to note the member arrived on `IDevice` rather than `IStreamingDevice`.
- The `RecordingTimestampProcessor` test fake gains `HasTimestampFrequency`, forwarded to the inner processor.

**The classification** (`SerialStreamingDevice.cs`):
- New arms for `NotFound`, `InUse`, `AccessDenied` -> `Warning`.
- `SerialPortConnectFailure.Unknown` deliberately **not** handled, so it still reaches `Error`/Sentry. Unknown means Core saw a serial-open failure it could not classify, which is worth investigating.
- The old `FileNotFoundException` / `UnauthorizedAccessException` arms are **kept** as a fallback for any path that still surfaces the raw platform exception rather than Core's translated one.

Per Core's own investigation, matching on `Reason` is also the first time this classification can be correct at all: a missing port and a busy port throw byte-identical exceptions on macOS, the inner message is not stable across processes for the same port, and the inner `HResult` is a stale errno. The old arms were matching on platform-specific accidents.

## Tests

Added to `SerialStreamingDeviceLogConnectFailureTests`:
- Each of the three classified reasons logs `Warning` and never `Error` (both `Error` overloads verified - the message-only one synthesizes an `AppLogErrorException` and would also reach Sentry).
- `Unknown` still logs `Error` - the guard in the other direction, so "downgrade everything" cannot pass.
- `SerialPortConnectException_DerivesFromIoException_NotTheLegacyPlatformTypes` asserts the hierarchy fact that makes the arms necessary.

That last one is deliberate. This is the **third** time Core changed a thrown type and a desktop classification arm silently stopped matching (#775, #779, #801) - each time discovered from Sentry volume rather than from a red test. It is a tripwire for the next one.

## Verification

- `dotnet build -tl:on` - clean, zero warnings
- `dotnet test --filter "TestCategory!=Ui&FullyQualifiedName!~WindowsFirewallWrapperTests" -tl:on` - **871/871 pass**

Notably the bump is behaviorally clean: unlike 1.1.1 -> 1.3.0, no existing test changed meaning. This is despite Core splitting `DaqifiStreamingDevice` and `FirmwareUpdateService` into collaborators and extracting `SdCardOperations`.

Bench pass to follow on this branch - the riskiest behavioral delta is Core's new `SerialStreamTransport` drop detection (write-probe plus port-presence poll), which is not covered by unit tests here.

## What this unblocks

#803, #804, #805, #679, #716, #752 (Stage 3), #704, #66. See the index comment on #802.

## Note for reviewers

`SerialPortConnectFailure.Unknown` reaching Sentry is a judgment call, not an oversight. If routine unclassified serial failures show up in Sentry volume after this ships, downgrading it is a one-line change - but starting permissive would hide a real signal.

### DIFF
--- a/Daqifi.Desktop.Test/Device/SerialStreamingDeviceLogConnectFailureTests.cs
+++ b/Daqifi.Desktop.Test/Device/SerialStreamingDeviceLogConnectFailureTests.cs
@@ -1,3 +1,5 @@
+using System.IO;
+using Daqifi.Core.Communication.Transport;
 using Daqifi.Desktop.Common.Loggers;
 using Daqifi.Desktop.Device.SerialDevice;
 using Moq;
@@ -153,6 +155,65 @@ public class SerialStreamingDeviceLogConnectFailureTests
         // ...and via the exception-carrying overload specifically: the message-only Error(string)
         // synthesizes an AppLogErrorException, which would strand the real stack trace out of Sentry.
         logger.Verify(l => l.Error(It.IsAny<string>()), Times.Never);
+    }
+
+    [TestMethod]
+    [DataRow(SerialPortConnectFailure.NotFound)]
+    [DataRow(SerialPortConnectFailure.InUse)]
+    [DataRow(SerialPortConnectFailure.AccessDenied)]
+    public void LogConnectFailure_WithClassifiedSerialPortConnectFailure_LogsWarningAndNotError(
+        SerialPortConnectFailure reason)
+    {
+        // Arrange — Core 1.4.0 (daqifi-core#427) translates SerialPort.Open failures into this typed
+        // exception. It derives from IOException, so it matches NEITHER of the FileNotFoundException
+        // / UnauthorizedAccessException arms that used to catch a missing or busy port. Without the
+        // Reason arms, all three of these — the most common things a user hits — reach the default
+        // Error path and are captured to Sentry as app bugs (issue #801).
+        var logger = new Mock<IAppLogger>();
+        var device = new TestableSerialStreamingDevice("COM_TEST_801", logger.Object);
+        var ex = new SerialPortConnectException("COM_TEST_801", reason, $"Serial open failed: {reason}");
+
+        // Act
+        device.ExposedLogConnectFailure(ex);
+
+        // Assert
+        logger.Verify(l => l.Warning(ex, It.IsAny<string>()), Times.Once);
+        logger.Verify(l => l.Error(It.IsAny<Exception>(), It.IsAny<string>()), Times.Never);
+        logger.Verify(l => l.Error(It.IsAny<string>()), Times.Never);
+    }
+
+    [TestMethod]
+    public void LogConnectFailure_WithUnknownSerialPortConnectFailure_LogsError()
+    {
+        // Arrange — the guard in the other direction. SerialPortConnectFailure.Unknown means Core saw
+        // a serial-open failure it could NOT classify, which is worth investigating, so it must keep
+        // reaching the Error/Sentry path rather than being blanket-downgraded with its siblings.
+        var logger = new Mock<IAppLogger>();
+        var device = new TestableSerialStreamingDevice("COM_TEST_801", logger.Object);
+        var ex = new SerialPortConnectException(
+            "COM_TEST_801", SerialPortConnectFailure.Unknown, "Serial open failed for an unknown reason");
+
+        // Act
+        device.ExposedLogConnectFailure(ex);
+
+        // Assert
+        logger.Verify(l => l.Error(ex, It.IsAny<string>()), Times.Once);
+        logger.Verify(l => l.Warning(It.IsAny<Exception>(), It.IsAny<string>()), Times.Never);
+        logger.Verify(l => l.Error(It.IsAny<string>()), Times.Never);
+    }
+
+    [TestMethod]
+    public void SerialPortConnectException_DerivesFromIoException_NotTheLegacyPlatformTypes()
+    {
+        // The structural fact that makes the arms above necessary, asserted directly so a future Core
+        // change to the hierarchy fails here with an explanatory name rather than silently re-routing
+        // routine connect failures to Sentry. This is the third time Core changed a thrown type and a
+        // classification arm quietly stopped matching (#775, #779, #801); this test is the tripwire.
+        var ex = new SerialPortConnectException("COM_TEST_801", SerialPortConnectFailure.NotFound, "x");
+
+        Assert.IsInstanceOfType<IOException>(ex);
+        Assert.IsNotInstanceOfType<FileNotFoundException>(ex);
+        Assert.IsNotInstanceOfType<UnauthorizedAccessException>(ex);
     }
 
     /// <summary>
--- a/Daqifi.Desktop.Test/Device/TimestampProcessorSerializationTests.cs
+++ b/Daqifi.Desktop.Test/Device/TimestampProcessorSerializationTests.cs
@@ -424,6 +424,12 @@ public class TimestampProcessorSerializationTests : IDisposable
 
         public double GetTickPeriod(string deviceId) => inner.GetTickPeriod(deviceId);
 
+        /// <summary>
+        /// Added to <see cref="ITimestampProcessor"/> in Core 1.4.0. Forwarded, not recorded: these
+        /// tests observe the apply/reset ordering, and this is a pure query.
+        /// </summary>
+        public bool HasTimestampFrequency(string deviceId) => inner.HasTimestampFrequency(deviceId);
+
         public TimestampResult ProcessTimestamp(string deviceId, uint deviceTimestamp)
         {
             Record(PROCESS_TIMESTAMP_CALL);
--- a/Daqifi.Desktop/Daqifi.Desktop.csproj
+++ b/Daqifi.Desktop/Daqifi.Desktop.csproj
@@ -38,7 +38,7 @@
 	</ItemGroup>
 	<ItemGroup>
 		<PackageReference Include="CommunityToolkit.Mvvm" Version="8.4.2" />
-		<PackageReference Include="Daqifi.Core" Version="1.3.0" />
+		<PackageReference Include="Daqifi.Core" Version="1.4.0" />
 		<PackageReference Include="EFCore.BulkExtensions.Sqlite" Version="10.0.1" />
 		<PackageReference Include="MahApps.Metro" Version="2.4.11" />
 		<PackageReference Include="MahApps.Metro.IconPacks.Material" Version="6.2.1" />
--- a/Daqifi.Desktop/Device/Firmware/BootloaderSessionStreamingDeviceAdapter.cs
+++ b/Daqifi.Desktop/Device/Firmware/BootloaderSessionStreamingDeviceAdapter.cs
@@ -2,6 +2,7 @@ using System.Net;
 using Daqifi.Core.Channel;
 using Daqifi.Core.Communication.Messages;
 using CoreConnectionStatus = Daqifi.Core.Device.ConnectionStatus;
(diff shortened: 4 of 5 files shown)


tylerkron:
## Bench results — real hardware, Nyquist on COM3

Ran an A/B against the physical device using a throwaway console app on the Core NuGet directly (no WPF, no FlaUI — avoids the CPU-contention flakes and isolates the transport). Identical scenario on both Core versions: connect, init, 20s idle, enable 4 analog inputs, stream 20s at 100 Hz, stop, restart 5s.

| | Core 1.3.0 (control) | Core 1.4.0 (this PR) |
|---|---|---|
| Init | 32 channels, 42 MHz clock | 32 channels, 42 MHz clock |
| Spurious status events, idle | 0 | 0 |
| Spurious status events, streaming | 0 | 0 |
| Frames, 20s stream | 1587 | 1586 |
| Frames, 5s restart | 396 | 395 |
| Background errors | n/a (no event in 1.3.0) | 0 |

### The risk this was testing: no false drops

1.4.0's `SerialStreamTransport` gained drop detection (write-probe plus port-presence poll) that can declare `ConnectionStatus.Lost`. A false positive would tear down a healthy device. **Zero spurious status events across 20s idle and 20s streaming** — the idle phase is the important one, since that is where the presence poll runs with no I/O to piggyback on. Status timeline was exactly `Connecting -> Connected -> Disconnected`.

`ErrorOccurred` also fired zero times, so nothing is failing silently in the background.

### Throughput is unchanged — the shortfall is pre-existing

Both versions land within **one frame** of each other (1586 vs 1587). The observed ~79 messages/sec at a requested 100 Hz is therefore **not** introduced by this bump; it reproduces identically on the version we ship today.

Worth noting it is not necessarily a defect either: this counts `StreamMessageReceived` **messages**, and firmware can batch multiple sample sets per protobuf message (see #449). Fewer messages than samples is the expected shape if batching is active. Attributing it properly is out of scope here — the point for this PR is that it is a constant across the bump.

### Bonus: Core's frame gate works on real hardware

Relevant to #679, which deletes desktop's equivalent guard. Core's `StreamFrameGate` caught both frame types during the 1.4.0 run:

```
DISCARD [PartialAnalogFrame]  ts=2806686261  analog=1/4   <- at stream start
DISCARD [StaleLeftoverFrame]  ts=3473519994  analog=1/4   <- at stream restart
```

The second one is exactly the #573 scenario — the device holding the previous session's final frame and emitting it first on resume. Core caught it unprompted, on this device, at the version this PR pins. That is the empirical evidence #679 wanted, obtained without downgrading the bench firmware.

tylerkron:
## Qodo review response

**Finding: test method names contain underscores (compliance ID 244817) — disagreed, not changed.**

The three new test methods (`LogConnectFailure_WithClassifiedSerialPortConnectFailure_LogsWarningAndNotError`, `LogConnectFailure_WithUnknownSerialPortConnectFailure_LogsError`, `SerialPortConnectException_DerivesFromIoException_NotTheLegacyPlatformTypes`) deliberately follow the `Method_Scenario_ExpectedResult` convention used by every other test in this file and across `Daqifi.Desktop.Test`. CA1707 is suppressed for the test project for exactly this reason.

Qodo itself scored this `● Weak` and cited the precedent: the same suggestion was raised and rejected in #235, #748, and #793. Renaming these three would make them inconsistent with the five existing methods in the same class they sit alongside.

No other findings were raised. CI is green (`build-and-test`, `validate-pr-title`), and bench results against real hardware are in the comment above — including an A/B against Core 1.3.0 confirming the bump is behaviorally neutral on streaming throughput and produces no false transport drops.

Requesting another pass below.

tylerkron:
Re-review came back with **no new findings** — the same single `● Weak` test-naming item, which Qodo itself annotates as previously rejected in #235, #748, and #793. Position unchanged (see above).

This PR is ready for human review from my side: CI green, 871/871 unit tests, and an A/B bench pass against real hardware confirming the bump is behaviorally neutral.

---

Unrelated process note for whoever picks this up: Qodo rule **244817 ("Enforce PascalCase for method names")** has now produced the same false positive on at least four PRs, because `Method_Scenario_ExpectedResult` is the house convention for tests and CA1707 is suppressed for the test project. Scoping that rule to exclude `**/*.Test/**` would stop it burning a review cycle every time. Filing that is outside this PR's scope, but it is cheap and recurring.

